PR #12, opened 2022-03-23
Fix session filenames stored in ~/.meticulous

leafty:
* ':' is not a valid character for filenames in Windows, causing issues
for replay


Gabriel-H (approved):
LGTM, 🙌

PR #78, opened 2022-08-08
Rename replay in commands

alexivanov:


packages/cli/src/api/replay.api.ts
```diff
@@ -127,7 +127,7 @@ export const postScreenshotDiffStats: (
 export const getReplayUrl = (replay: any) => {
   const organizationName = encodeURIComponent(replay.project.organization.name);
   const projectName = encodeURIComponent(replay.project.name);
-  const replayUrl = `https://app.meticulous.ai/projects/${organizationName}/${projectName}/replays/${replay.id}`;
+  const replayUrl = `https://app.meticulous.ai/projects/${organizationName}/${projectName}/simulations/${replay.id}`;
```
leafty, line 130: Can we split out the URL changes?

leafty (changes requested):
Let's split the URL changes in another PR. Otherwise LGTM

packages/cli/src/api/replay.api.ts
alexivanov: good shout 👍 https://github.com/alwaysmeticulous/meticulous-sdk/pull/78/commits/554a755f9c82e64d74f1ac26d8ec64c05eb3b58f removes the changes from this PR. I think we can ship this is an a following release.

leafty (approved):
LGTM

packages/cli/src/api/replay.api.ts
leafty: nice, we can ship this first then :+1:

PR #79, opened 2022-08-08
Close puppeteer after recording has finished, i.e. page has been closed

alexivanov:


leafty (approved):
LGTM

:thinking: interesting, chromium would close for me when closing all the tabs. it seems that it's not behaving the same way on macos.

PR #82, opened 2022-08-18
Replay resize events

alexivanov:
### Changes
This PR adds support for handling window resize events.

For simulation we expose a `setViewport` puppeteer function which is invoked as part of "resize" event simulation.


leafty (approved):
LGTM; worked when I did record/simulate.

PR #84, opened 2022-08-22
make base snippets URL configurable via an env variable

alexivanov:
This will allow changing environment, for example
changing from production to staging / development,
easier.

### DIFF
--- a/packages/cli/src/config/snippets.ts
+++ b/packages/cli/src/config/snippets.ts
@@ -1,9 +1,5 @@
 const BASE_SNIPPETS_URL = "https://snippet.meticulous.ai/";
 
 export const getSnippetsBaseUrl = (): string => {
-  const baseURL =
-    process.env["METICULOUS_SNIPPETS_BASE_URL"] || BASE_SNIPPETS_URL;
-
-  // Append trailing slack if missing from the provided base URL.
-  return baseURL.endsWith("/") ? baseURL : baseURL + "/";
+  return process.env["METICULOUS_SNIPPETS_BASE_URL"] || BASE_SNIPPETS_URL;
 };
--- a/packages/cli/src/config/snippets.ts
+++ b/packages/cli/src/config/snippets.ts
@@ -1,0 +1,9 @@
+const BASE_SNIPPETS_URL = "https://snippet.meticulous.ai/";
+
+export const getSnippetsBaseUrl = (): string => {
+  const baseURL =
+    process.env["METICULOUS_SNIPPETS_BASE_URL"] || BASE_SNIPPETS_URL;
+
+  // Append trailing slack if missing from the provided base URL.
+  return baseURL.endsWith("/") ? baseURL : baseURL + "/";
+};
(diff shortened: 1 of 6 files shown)


packages/cli/src/config/snippets.ts
```diff
@@ -0,0 +1,9 @@
+const BASE_SNIPPETS_URL = "https://snippet.meticulous.ai/";
+
+export const getSnippetsBaseUrl = (): string => {
+  const baseURL =
+    process.env["METICULOUS_SNIPPETS_BASE_URL"] || BASE_SNIPPETS_URL;
+
+  // Append trailing slack if missing from the provided base URL.
+  return baseURL.endsWith("/") ? baseURL : baseURL + "/";
```
leafty, line 8: You don't need this if you use `new URL` to construct them. See other comment.

packages/cli/src/local-data/replay-assets.ts
```diff
@@ -42,10 +43,9 @@ export const saveAssetMetadata: (
   await writeFile(assetsFile, JSON.stringify(assetMetadata, null, 2));
 };
 
-export const fetchAsset: (fetchUrl: string) => Promise<string> = async (
-  fetchUrl
-) => {
+export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
+  const fetchUrl = `${getSnippetsBaseUrl()}${path}`;
```
leafty, line 48: You can use `new URL` here:

```js
> (new URL('some/foo/bar.js', 'http://example.com')).href
'http://example.com/some/foo/bar.js'
> (new URL('some/foo/bar.js', 'http://example.com/')).href
'http://example.com/some/foo/bar.js'
```

packages/cli/src/config/snippets.ts
alexivanov: 💯 good shout! https://github.com/alwaysmeticulous/meticulous-sdk/pull/84/commits/449da2ca083d48e50b515ffb2a55357a2bb82c35 does this

packages/cli/src/config/snippets.ts
```diff
@@ -0,0 +1,5 @@
+const BASE_SNIPPETS_URL = "https://snippet.meticulous.ai/";
+
+export const getSnippetsBaseUrl = (): string => {
+  return process.env["METICULOUS_SNIPPETS_BASE_URL"] || BASE_SNIPPETS_URL;
+};
```
leafty, line 5: To validate `METICULOUS_SNIPPETS_BASE_URL`, I would do:

```suggestion
export const getSnippetsBaseUrl = (): string => {
  try {
    return new URL(process.env["METICULOUS_SNIPPETS_BASE_URL"] || BASE_SNIPPETS_URL).href;
  } catch (error) {
    if (error instanceof TypeError) {
      return BASE_SNIPPETS_URL;
    }
    throw error;
  }
};
```

And maybe also memoize the value? Not needed to merge though :smile:

leafty (approved):
LGTM, just check the comment

packages/cli/src/config/snippets.ts
alexivanov: I'd err on the side of loudly erroring rather than swallowing the error here and defaulting, https://github.com/alwaysmeticulous/meticulous-sdk/pull/84/commits/384411ab2396db7f9f042461a60833152d2743f9 makes the validation error bit clear though.

leafty: That works too! :ok_hand:

PR #87, opened 2022-08-23
Use start URL from session metadata

alexivanov:
### Motivation

The session metadata `startURL` is now considered the canonical way of getting the first URL of a session so during replay we should use that.

Most session now should have a startURL defined in metadata but keeping the old event-stream startURL logic in place for backward-compatibility, we can drop it in the future.


leafty (approved):
LGTM

PR #88, opened 2022-08-24
Change screenshot log statement to debug level

Gabriel-H:
We should display a single next action for the user to take and currently the link into the Meticulous dashboard provides this.


leafty (approved):
LGTM

PR #90, opened 2022-08-25
set screenshot flag to true by default in simulate command

alexivanov:
In most cases users want a final state screenshot during simulation


leafty (approved):
Yes, `false` sounded good when starting to do this way for Meticulous. Now it should be on by default.

PR #91, opened 2022-08-25
drop screenshot argument of create-test command

alexivanov:
Tests are based on visual difference of the final simulation state so screenshots are a requirement for creating a test at the moment.


leafty (approved):
OK, this makes sense :D

PR #95, opened 2022-09-06
Run lint in CI

alexivanov:


.github/workflows/build.yaml
```diff
@@ -0,0 +1,29 @@
...
+          node-version: "16"
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --non-interactive
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Build packages
+        run: yarn run turbo run build
```
leafty, line 29: Note: We should define a top-level `build` script

leafty (approved):
LGTM :ok_hand:

.github/workflows/build.yaml
alexivanov: 👍 yep good shout. Will do this in a separate RP

PR #96, opened 2022-09-06
avoid defining empty interface for bootstrap command options

alexivanov:


alexivanov:
Looks like this is fixed on `main` already 🙃

leafty:
Ah yeah, but we should probably lint on PRs

PR #97, opened 2022-09-09
Run all tests: add `--useCache` option

leafty:
* the `--useCache` option will not re-run passing tests for the same commit

### DIFF
--- a/packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
+++ b/packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
@@ -34,7 +34,7 @@ interface Options {
   parallelize?: boolean | null | undefined;
   parallelTasks?: number | null | undefined;
   deflake: boolean;
-  cache: boolean;
+  useCache: boolean;
 }
 
 const handler: (options: Options) => Promise<void> = async ({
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
```diff
@@ -66,6 +70,11 @@ const handler: (options: Options) => Promise<void> = async ({
   const commitSha = (await getCommitSha(commitSha_)) || "unknown";
   const meticulousSha = await getMeticulousVersion();
 
+  const cachedTestRunResults =
+    useCache && commitSha
```
alexivanov, line 74: I think `commitSha` will always have a truthy value here based on the [fallback value](https://github.com/alwaysmeticulous/meticulous-sdk/pull/97/files#diff-087c4a7f307a730c3bbf0be5c05e89e3d8e5e0594702ac66f703ed33b29d3325R70).

Maybe in cases where `useCache` is set but command is ran outside of git repo we should error loudly too?

leafty: oh, thanks for spotting that

PR #99, opened 2022-09-26
key assets metadata by file name rather than fetch URL location

alexivanov:
## Motivation

Assets locations change per environment and we want to re-fetch any
assets which differ in case of an environment change. Keying the
asset metadata by the output file name ensures a correct etag
comparison (comparing etag of the local file vs the remote origin)
in case of an environment change.

### DIFF
--- a/packages/cli/src/local-data/replay-assets.ts
+++ b/packages/cli/src/local-data/replay-assets.ts
@@ -10,7 +10,7 @@ import { getSnippetsBaseUrl } from "../config/snippets";
 
 export interface AssetMetadataItem {
   fileName: string;
-  etag: string;
+  etag: string | null;
   fetchUrl: string;
 }
 
@@ -54,7 +54,7 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const assetFileName = basename(new URL(fetchUrl).pathname);
 
   const assetMetadata = await loadAssetMetadata();
-  const etag = (await axios.head(fetchUrl)).headers["etag"] || "";
+  const etag = (await axios.head(fetchUrl)).headers["etag"] || null;
 
   const entry = assetMetadata.assets.find(
     (item) => item.fileName === assetFileName
@@ -62,7 +62,7 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const fileName = assetFileName;
   const filePath = join(await getCreateAssetsDir(), fileName);
 
-  if (entry && etag === entry.etag) {
+  if (entry && etag && etag === entry.etag) {
     logger.debug(`${fetchUrl} already present`);
     return filePath;
   }
--- a/packages/cli/src/local-data/replay-assets.ts
+++ b/packages/cli/src/local-data/replay-assets.ts
@@ -18,10 +18,17 @@ export interface AssetMetadata {
   assets: AssetMetadataItem[];
 }
 
-export const loadAssetMetadata: () => Promise<AssetMetadata> = async () => {
-  const assetsDir = join(getMeticulousLocalDataDir(), "assets");
+const ASSETS_FOLDER_NAME = "assets";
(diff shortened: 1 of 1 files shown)


packages/cli/src/local-data/replay-assets.ts
```diff
@@ -10,18 +10,25 @@ import { getSnippetsBaseUrl } from "../config/snippets";
...
 export interface AssetMetadata {
   assets: AssetMetadataItem[];
 }
 
-export const loadAssetMetadata: () => Promise<AssetMetadata> = async () => {
-  const assetsDir = join(getMeticulousLocalDataDir(), "assets");
+const ASSETS_FOLDER_NAME = "assets";
+const ASSET_METADATA_FILE_NAME = "assets.json";
+
+export const getCreateAssetsDir: () => Promise<string> = async () => {
```
leafty, line 24:
```suggestion
export const getOrCreateAssetsDir: () => Promise<string> = async () => {
```

packages/cli/src/local-data/replay-assets.ts
```diff
@@ -36,29 +43,26 @@ export const loadAssetMetadata: () => Promise<AssetMetadata> = async () => {
...
 export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
   const fetchUrl = new URL(path, getSnippetsBaseUrl()).href;
-
-  const assetsDir = join(getMeticulousLocalDataDir(), "assets");
+  const assetFileName = basename(new URL(fetchUrl).pathname);
 
   const assetMetadata = await loadAssetMetadata();
-  const etag = (await axios.head(fetchUrl)).headers["etag"] || "";
+  const etag = (await axios.head(fetchUrl)).headers["etag"] || null;
```
leafty, line 57: Is there anything against doing this?
```suggestion
  const etag = (await axios.head(fetchUrl)).headers["etag"] || "";
```

Then the type can remain `string`.

packages/cli/src/local-data/replay-assets.ts
```diff
@@ -36,29 +43,26 @@ export const loadAssetMetadata: () => Promise<AssetMetadata> = async () => {
...
-  const entry = assetMetadata.assets.find((item) => item.fetchUrl === fetchUrl);
-  const fileName = entry
-    ? entry.fileName
-    : basename(new URL(fetchUrl).pathname);
-  const filePath = join(assetsDir, fileName);
+  const entry = assetMetadata.assets.find(
+    (item) => item.fileName === assetFileName
+  );
+  const fileName = assetFileName;
+  const filePath = join(await getCreateAssetsDir(), fileName);
```
leafty, line 63:
```suggestion
  const filePath = join(await getCreateAssetsDir(), assetFileName);
```

packages/cli/src/local-data/replay-assets.ts
alexivanov: https://github.com/alwaysmeticulous/meticulous-sdk/pull/99/commits/14c14f45010c1fe4aecf655bdac4b6008b89577f

packages/cli/src/local-data/replay-assets.ts
alexivanov: good shout https://github.com/alwaysmeticulous/meticulous-sdk/pull/99/commits/5fb3dfe871276f07c52975c1c3fab6a187fbb7c1

packages/cli/src/local-data/replay-assets.ts
alexivanov: I feel like `null` provides more clarity that the eTag wasn't set rather than an empty string (which might also be a valid value?)

leafty: I don't think we should treat `""` as a valid `etag`, it does not make sense

leafty (approved):
LGTM

PR #100, opened 2022-09-30
Snapshot assets when running replays, and allow re-executing simulations against snapshotted assets

Que3216:
Allows the user to append a `--simulationIdForAssets <simulation id>` to their simulate command, and it'll download the assets captured during a prior run of the specified simulation, and use those to run the new simulation against. Useful for debugging simulations (without having to find the specific commit they were run against, get the code running locally), and for us to test how changes to our algorithms affect prior simulations.


packages/cli/package.json
```diff
@@ -26,11 +26,14 @@
   "dependencies": {
     "@alwaysmeticulous/common": "^1.7.6",
     "@sentry/node": "^7.2.0",
+    "@types/express": "^4.17.14",
```
leafty, line 29: Move to a dev dependency

packages/cli/package.json
```diff
@@ -26,11 +26,14 @@
...
     "adm-zip": "^0.5.9",
     "archiver": "^5.3.1",
     "axios": "^0.27.2",
     "chalk": "^4.1.2",
     "cosmiconfig": "^7.0.1",
+    "express": "^4.18.1",
```
leafty, line 35: It feels a bit wrong to make the sdk depend on `express` :thinking:

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -84,6 +87,9 @@ export const replayCommandHandler: (
   cookies,
   cookiesFile,
 }) => {
+  let closeServerIfRunning = () => {
```
leafty, line 90: `let` :scream: 

Also, no need to define a noop function. If `closeServerIfRunning` has type `Function | null | undefined`, you can call it with `closeServerIfRunning?.();` (will call the function if it is not null or undefined).

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -98,15 +104,25 @@ export const replayCommandHandler: (
...
 
-  // 3. Load replay assets
+  // 3. If simulationIdForAssets specified then download assets & spin up local server
+  if (simulationIdForAssets) {
+    const result = await serveAssetsFromSimulation(
+      client,
+      simulationIdForAssets
+    );
+    appUrl = result.url;
+    closeServerIfRunning = result.closeServer;
```
leafty, line 114: Not a fan of mutable variables if we could avoid that :thinking:

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -184,6 +200,7 @@ export const replayCommandHandler: (
 
   await eventsFinishedPromise;
   await writesFinishedPromise;
+  closeServerIfRunning();
```
leafty, line 203: Could be `closeServerIfRunning?.();`

packages/cli/src/commands/replay/replay.command.ts
Que3216: Me neither! I could do:

```ts
const server = simulationIdForAssets ? await serveAssetsFromSimulation(
      client,
      simulationIdForAssets
    ) : undefined;
const urlToRunSimulationAgainst = server ? server.url : appUrl ;

[...] (and replace refs to appUrl with urlToRunSimulationAgainst)

server?.closeServer?.();
```

packages/cli/package.json
Que3216: I think depending on something to serve assets is inevitable if we want users to be able to copy and paste a command to re-run a replay locally for debugging, without having to spin up a server for the given commit. It's not too large though considering that this is a CLI, and dev connections are normally pretty fast: https://bundlephobia.com/package/express@4.18.1.

packages/cli/package.json
Que3216: Ooops yes! I'll update, good catch!

packages/cli/src/commands/replay/replay.command.ts
Que3216: 👍 I'll update

packages/cli/src/commands/replay/replay.command.ts
Que3216: I've refactored, I think this is nicer

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -184,6 +192,7 @@ export const replayCommandHandler: (
 
   await eventsFinishedPromise;
   await writesFinishedPromise;
+  server?.closeServer?.();
```
leafty, line 195:
```suggestion
  server?.closeServer();
```

I think, since the method exists when the server does.

leafty (approved):
LGTM

PR #101, opened 2022-10-03
Navigate directly to recording page docs when browser opens

Gabriel-H:
This navigates to the recording-a-test docs url as the first page when launching the record command. We do a hard-code check for the recording-a-page doc page, to avoid injecting the snippet on this page and to avoid creating a extra dummy recording


leafty:
oh sorry, I already merged it :)

PR #104, opened 2022-10-03
Fallback to loading API token from personal config file

Que3216:


packages/cli/src/api/client.ts
```diff
@@ -11,6 +13,13 @@ export const createClient: (options: ClientOptions) => AxiosInstance = ({
   apiToken: apiToken_,
 }) => {
   const apiToken = getApiToken(apiToken_);
+  if (!apiToken) {
+    const logger = log.getLogger(METICULOUS_LOGGER_NAME);
+    logger.error(
```
Que3216, line 18: I checked through all the commands, and I think they all require an API token (all hit BE endpoints that require auth)?

leafty:
One thing which I'm not sure about is that this is potentially a bad security practice (storing tokens clear on disk). :thinking:

Que3216:
Yeah it's a good flag -- we may want to add a more secure option in future, such as loading from keychain or 1password, or loading from encrypted file and prompting for de-encrypt password.

PR #107, opened 2022-10-04
Allow specifying a different replay id to use assets from, and wire through snapshotted assets options to parallelise code paths

Que3216:


leafty (commented):
Code is OK(-ish) but it feels like cardboard and tape we should replace with internal, on-demand infra to serve frontend applications.

PR #111, opened 2022-10-06
Update replayer package implementation to Replay Next

alexivanov:
A new major replayer package version which moves some of the control plane logic to the node puppeteer process. Moving the logic outside of the window environment enables replay of sessions with navigation events.


leafty (approved):
LGTM :champagne: :tada: :trophy:

PR #112, opened 2022-10-07
Create a package for loading the Meticulous recording snippet

alexivanov:


packages/recorder-loader/src/loader.ts
```diff
@@ -0,0 +1,69 @@
...
+      try {
+        window.__meticulous?.initialiseRecorder();
+      } catch (Error) {
+        console.error("Meticulous recorder failed to initialise.");
+      }
+
+      resolve();
+    };
+    script.onerror = () => {
+      resolve();
```
leafty, line 64: Would it not be better to `reject()` in this case?

leafty (approved):
LGTM

packages/recorder-loader/src/loader.ts
```diff
@@ -0,0 +1,69 @@
...
+          "Meticulous snippet abandoned due to max blocking time reached."
+        );
+        // At this point the promise has already resolved.
+        return;
+      }
+
+      try {
+        window.__meticulous?.initialiseRecorder();
+      } catch (Error) {
+        reject("Meticulous recorder failed to initialise.");
```
Que3216, line 58: Side note (don't need to make this change) -- if we want to make it even fewer lines of code for our users we could make these resolves rather than rejects, and rename to 'tryToLoadAndStartRecorder':

```
async function startApp() {
     await tryToLoadAndStartRecorder({
            projectId: '<project ID>',
     })
    ReactDOM.render(component, document.getElementById('root'))
}


startApp();
```

packages/recorder-loader/src/loader.ts
```diff
@@ -0,0 +1,69 @@
...
+        console.debug(
+          "Meticulous snippet abandoned due to max blocking time reached."
+        );
+        // At this point the promise has already resolved.
+        return;
+      }
+
+      try {
+        window.__meticulous?.initialiseRecorder();
+      } catch (Error) {
```
Que3216, line 57: We silently swallow the error here and in script.onerror below (we don't take an error parameter). We don't want to pollute customers logs, but the silent swallowing could make debugging hard, and also means that in your example code, there's not much point the customer logging the error:

```
async function startApp() {
    // Start the Meticulous recorder before you initialise your app.
    try {
        await loadAndStartRecorder({
            projectId: '<project ID>',
        })
    } catch (err) {
        console.error(`Meticulous failed to initialise ${err}`)
    }
    // Initalise app after Meticulous' snippet, e.g:
    ReactDOM.render(component, document.getElementById('root'))
}
```

Should we propagate the error as part of our reject call?

We could also always log a debug message if loading meticulous fails.

PR #113, opened 2022-10-11
[Asset snapshotting] Correctly handle all cases of serving HTML files

Que3216:


packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -1,5 +1,5 @@
 import { mkdir, writeFile } from "fs/promises";
-import { dirname, join } from "path";
+import path, { dirname, extname, join } from "path";
```
leafty, line 2: Apparently imports should be optimized here.

leafty (approved):
LGTM

Note: it feels that this code is doing exactly what a server dedicated to serving single-page apps would do.

packages/replayer/src/assets/snapshot-assets.ts
Que3216: Ah yes! Not sure how that slipped in. I'll fix.

Que3216:
```suggestion
import { dirname, extname, join } from "path";
```

PR #114, opened 2022-10-12
Add serve command

Que3216:
Add command to spin up a localhost server to serve the assets that were snapshotted when running a particular replay.

I added this mainly for my own usage in debugging, but I think it's useful and worth merging. What do you think?

Downside would be more commands get listed when you run `npx meticulous`, which could make it more confusing for users to learn how to use the product. (ideal thing would be to start grouping the commands into core and advanced, not sure that's possible)

### DIFF
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -9,6 +9,7 @@ import { record } from "./commands/record/record.command";
 import { replay } from "./commands/replay/replay.command";
 import { runAllTests } from "./commands/run-all-tests/run-all-tests.command";
 import { screenshotDiff } from "./commands/screenshot-diff/screenshot-diff.command";
+import { serve } from "./commands/serve/serve.command";
 import { showProject } from "./commands/show-project/show-project.command";
 import { updateTests } from "./commands/update-tests/update-tests.command";
 import { uploadBuild } from "./commands/upload-build/upload-build.command";
@@ -44,6 +45,7 @@ export const main: () => void = () => {
     .command(showProject)
     .command(updateTests)
     .command(uploadBuild)
+    .command(serve)
     .help()
     .strict()
     .demandCommand()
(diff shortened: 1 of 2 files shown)


packages/cli/src/main.ts
```diff
@@ -44,6 +45,7 @@ export const main: () => void = () => {
     .command(showProject)
     .command(updateTests)
     .command(uploadBuild)
+    .command(serve)
```
alexivanov, line 48: Looks like this will hide it from the help menu
```suggestion
    .command("serve", false, serve)
```

alexivanov (approved):

leafty (approved):
LGTM

I think maybe this type of utilities should not be added to this repository.

packages/cli/src/main.ts
Que3216: Oh nice, I'll try that

Que3216:
> I think maybe this type of utilities should not be added to this repository.

You're worried about more bytes for customers to download, or making the CLI help screen more confusing by having lots of junk commands?

We could have a separate debugging cli. Though I think if Alex's hiding suggestion works though then should be OK for commands that are lightweight like this / better than overhead of maintaining two CLIs and sharing the code amongst them. So will test that first.

leafty:
> > I think maybe this type of utilities should not be added to this repository.
> 
> You're worried about more bytes for customers to download, or making the CLI help screen more confusing by having lots of junk commands?
> 
> We could have a separate debugging cli. Though I think if Alex's hiding suggestion works though then should be OK for commands that are lightweight like this / better than overhead of maintaining two CLIs and sharing the code amongst them. So will test that first.

Simply put, this repository is code which we ship to users and is ISC-licensed :) It is easier for us to make internal utilities we move here when needed than try to do the opposite.

packages/cli/src/main.ts
Que3216: It works -- thanks!

PR #115, opened 2022-10-12
Add accelerate flag

Que3216:
TODO in follow up PR:
 - [ ] Deduplicate some of the common interface, so don't have to repeat these type declarations for all the replay args so many times

### DIFF
--- a/packages/cli/src/commands/replay/replay.command.ts
+++ b/packages/cli/src/commands/replay/replay.command.ts
@@ -60,7 +60,7 @@ export interface ReplayCommandHandlerOptions {
   moveBeforeClick?: boolean | null | undefined;
   cookies?: Record<string, any>[];
   cookiesFile?: string | null | undefined;
-  accelerate?: boolean | null | undefined;
+  accelerate: boolean;
 }
 
 export const replayCommandHandler: (
@@ -191,7 +191,7 @@ export const replayCommandHandler: (
     moveBeforeClick: moveBeforeClick || false,
     cookies: cookies || null,
     cookiesFile: cookiesFile || "",
-    accelerate: accelerate ?? undefined,
+    accelerate,
   };
   await writeFile(
     join(tempDir, "replayEventsParams.json"),
--- a/packages/cli/src/commands/replay/replay.command.ts
+++ b/packages/cli/src/commands/replay/replay.command.ts
@@ -60,6 +60,7 @@ export interface ReplayCommandHandlerOptions {
   moveBeforeClick?: boolean | null | undefined;
   cookies?: Record<string, any>[];
   cookiesFile?: string | null | undefined;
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -189,6 +191,7 @@ export const replayCommandHandler: (
     moveBeforeClick: moveBeforeClick || false,
     cookies: cookies || null,
     cookiesFile: cookiesFile || "",
+    accelerate: accelerate ?? undefined,
```
leafty, line 194:
```suggestion
    accelerate: accelerate || false,
```

or 

```suggestion
    accelerate: accelerate ?? false,
```

packages/cli/src/parallel-tests/messages.types.ts
```diff
@@ -19,6 +19,7 @@ export interface InitMessage {
       padTime: boolean;
       shiftTime: boolean;
       networkStubbing: boolean;
+      accelerate: boolean | undefined;
```
leafty, line 22:
```suggestion
      accelerate: boolean;
```

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -37,6 +37,7 @@ export interface RunAllTestsInParallelOptions {
   parallelTasks: number | null | undefined;
   deflake: boolean;
   cachedTestRunResults: TestCaseResult[];
+  accelerate: boolean | undefined;
```
leafty, line 40:
```suggestion
  accelerate: boolean;
```

packages/common/src/types/replay.types.ts
```diff
@@ -38,6 +38,7 @@ export interface ReplayEventsOptions {
   moveBeforeClick: boolean;
   cookies: Record<string, any>[] | null;
   cookiesFile: string;
+  accelerate?: boolean | undefined;
```
leafty, line 41:
```suggestion
  accelerate: boolean;
```

packages/cli/src/parallel-tests/messages.types.ts
Que3216: I thought about doing this, but did the undefined thing instead to avoid having the default value of 'false' copy and pasted into too many places (think we want to keep our default value for acceleration the same between all commands, otherwise it'd be confusing). So we pass a boolean or undefined through most the levels, and then there's a single place in replayEvents which includes the logic of setting to default value of false if undefined.

Que3216: Update: actually I think we can get best of both worlds — I’ll add a shared const with default values for common command parameters, move the default application to the command level, and use that.

leafty (approved):
LGTM

PR #116, opened 2022-10-13
GitHub actions: update actions to latest version

leafty:
* Also, add yarn and build caching

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -17,12 +17,21 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Use Node.js LTS
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
-          node-version: "16"
+          node-version: 16
+          cache: yarn
+
+      - name: Cache turborepo output
+        uses: actions/cache@v3
+        with:
+          path: |
+            node_modules/.cache/turbo
+            packages/*/.turbo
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -17,12 +17,21 @@ jobs:
...
+          node-version: 16
+          cache: yarn
+
+      - name: Cache turborepo output
+        uses: actions/cache@v3
+        with:
+          path: |
+            node_modules/.cache/turbo
+            packages/*/.turbo
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
```
Que3216, line 34: Note that this will mean the size of the cache entry will grow with each commit that doesn't change the yarn.lock, but I don't think this will be an issue in practice since the yarn.lock will grow frequently enough (also presume turbo does some trimming of its cache occasionally).

Que3216: Also it looks like the cache isn't updated when there's a cache hit:

 "Cache hit occurred on the primary key Linux-02f1ec71e458f0fd9b2c3767e15a260f037f96ee6d2b32563230939abe456ea2, not saving cache."

(https://github.com/alwaysmeticulous/meticulous-sdk/actions/runs/3240966283/jobs/5312306603)

This means we'll still get some caching, but e.g. the third commit pushed with a given unique yarn.lock hash won't take advantage of build results from the build on the 2nd commit, only those from the 1st commit.

Que3216: I don't think this is an issue though (will get plenty of speed up still). Can always use turbo remote cache if we later find we still want more speed.

PR #118, opened 2022-10-14
Fix issues with snapshotting assets in case of cancelled requests

Que3216:


leafty (commented):
First quick pass: it seems to me the bug could be avoided by checking the response status. https://pptr.dev/api/puppeteer.httpresponse.ok
Having a failsafe timeout is still a good idea.

Que3216:
> First quick pass: it seems to me the bug could be avoided by checking the response status. https://pptr.dev/api/puppeteer.httpresponse.ok

Yeah that's what I assumed too. But when I printed out the responses that were causing an issue, they were 200 OK (which is not what I would have expected for a cancelled request). It's possible I made a mistake somewhere though -- did you manage to get it to work by only recorded ok() requests?

packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -14,29 +16,69 @@ const resourceTypesToSnapshot = new Set<ResourceType>([
...
+interface SnapshotAssetsOpts {
   assetsPath: string;
   baseUrl: string;
   assetSnapshots: AssetSnapshot[];
-}) => {
+}
+
+const TIMEOUT_FOR_FETCHING_ASSET = 5000;
+
+export const snapshotAssets = async (opts: SnapshotAssetsOpts) => {
```
Que3216, line 27:
```ts
export const snapshotAssets: (options: SnapshotAssetsOpts) => Promise<string> = async ({ assetSnapshots, ... }) => {
```

vs

```ts
export const snapshotAssets = async ({ assetSnapshots, ... }: SnapshotAssetsOpts): Promise<string> => {
```

Que3216: Deconstruct on next line

packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -68,3 +110,34 @@ function getFilePath(trimmedUrl: string, contentType: string | undefined) {
   }
   return trimmedUrl;
 }
+
+const withTimeout = <T>(
```
Que3216, line 114: `withTimeout(promise, Duration.fromObject({ seconds: 1 }), "Oh no")`

Que3216:
@leafty I've updated now to address style issues -- let me know what you think

packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -14,29 +17,72 @@ const resourceTypesToSnapshot = new Set<ResourceType>([
...
+    // seem to still have 200 status etc. in Puppeteer oddly). We ignore this, since we don't
+    // care about snapshotting assets for cancelled requests.
+    if ((error as any)?.name === "ProtocolError") {
+      logger.debug(
+        `ProtocolError when fetching snapshotted asset for URL ${snapshot.url}. Ignoring this, ` +
+          `because it's usually due to cancelled requests.`
+      );
+    } else {
+      throw error;
+    }
```
leafty, line 84: nitpick:
```suggestion
    if ((error as any)?.name === "ProtocolError") {
      logger.debug(
        `ProtocolError when fetching snapshotted asset for URL ${snapshot.url}. Ignoring this, ` +
          `because it's usually due to cancelled requests.`
      );
      return;
    }
    
    throw error;
```

packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -67,4 +116,37 @@ function getFilePath(trimmedUrl: string, contentType: string | undefined) {
...
+        timeoutDuration.toMillis()
+      );
+      promise.finally(() => clearTimeout(cancellationId));
+    }),
+  ]);
+};
+
+/**
+ * We only care to snapshot the first response for a given URL
+ */
```
leafty, line 140:
```suggestion
// We only care to snapshot the first response for a given URL
```

packages/replayer/src/replay.utils.ts
```diff
@@ -197,7 +197,7 @@ export const initializeReplayData: (options: {
       assetSnapshots.push({
         url: response.request().url(),
         contentType: response.headers()["content-type"],
```
leafty, line 199:
```suggestion
        contentType: response.headers()["content-type"] || "",
```

packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -48,13 +94,16 @@ export const snapshotAssets = async (opts: {
...
   const parsed = new URL(url);
   parsed.search = "";
   return parsed.toString();
-}
+};
 
-function getFilePath(trimmedUrl: string, contentType: string | undefined) {
+const getFilePath: (
+  trimmedUrl: string,
+  contentType: string | undefined
```
leafty, line 105:
```suggestion
  contentType: string
```
to avoid `undefined`.

packages/replayer/src/assets/snapshot-assets.ts
```diff
@@ -48,13 +94,16 @@ export const snapshotAssets = async (opts: {
...
+) => string = (trimmedUrl, contentType) => {
   const hasHTMLContentType =
     contentType !== undefined && contentType.indexOf("text/html") > -1;
```
leafty, line 108: We don't need to use `indexOf` here.
```suggestion
  const hasHTMLContentType = contentType.includes("text/html");
```

leafty (approved):
LGTM :tada: , just a few comments.

PR #127, opened 2022-10-21
acceleratePlayback -> useVirtualTime

Que3216:


leafty (approved):

packages/replayer/src/replayer.ts
```diff
@@ -170,7 +174,7 @@ export const replayEvents: ReplayEventsFn = async (options) => {
     logLevel,
     sessionData,
     moveBeforeClick: true,
-    acceleratePlayback: accelerate,
+    virtualTime: accelerate ? { enabled: false } : { enabled: true },
```
leafty, line 177: Unless I am going crazy:

```suggestion
    virtualTime: accelerate ? { enabled: true } : { enabled: false },
```

Que3216: No you're right! I don't know how I missed this.

PR #129, opened 2022-10-21
Smaller code coverage files

Que3216:


packages/replayer/src/output.utils.ts
```diff
@@ -77,13 +79,30 @@ const writeRawCoverageData: (options: {
...
     join(outputDir, "raw-coverage.json"),
-    JSON.stringify(coverageData, null, 2),
+    JSON.stringify(withoutMeticulousScriptCoverage, null, 2),
     "utf-8"
   );
 };
 
+const isMeticulousSnippetURL = (url: string) => {
+  try {
+    return new URL(url).origin === METICULOUS_SNIPPET_ORIGIN;
```
alexivanov, line 98: Could potentially use `getSnippetsBaseUrl` instead of redefining the snippet origin here - will probably require moving that util function to the common package though. Up to you if you want to do the refactoring as part of this.

alexivanov (approved):

packages/replayer/src/output.utils.ts
```diff
@@ -77,13 +79,30 @@ const writeRawCoverageData: (options: {
   outputDir: string;
   coverageData: CoverageEntry[];
 }) => Promise<void> = async ({ outputDir, coverageData }) => {
+  // We already snapshot the assets, so don't need to save the script contents
+  const withoutScriptText = coverageData.map(
+    ({ text: _text, ...rest }) => rest // eslint-disable-line @typescript-eslint/no-unused-vars
+  );
```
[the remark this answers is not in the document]
Que3216: Side note: could set ignoreRestSiblings to true

packages/replayer/src/output.utils.ts
Que3216: Ah yeah this is a good call / worth doing. I'll update.

PR #130, opened 2022-10-24
Add types to describe the `replayUserInteractions()` function

leafty:


packages/replayer/src/output.utils.ts
```diff
@@ -2,7 +2,7 @@ import { BASE_SNIPPETS_URL, SessionData } from "@alwaysmeticulous/common";
 import { mkdir, writeFile } from "fs/promises";
 import { join } from "path";
 import { ReplayData, ReplayMetadata } from "./replay.types";
-import { ReplayTimelineData } from "./timeline/timeline.types";
+import { ReplayTimelineData } from "@alwaysmeticulous/sdk-bundles-api";
```
Que3216, line 5: 🥳

packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts
```diff
@@ -1,3 +1,43 @@
...
+/** Returned when the recorded session has been fully replayed */
+export interface ReplayUserInteractionsResultFull {
+  length: "full";
+}
+
+/** Returned when the recorded session has been cut short during replay */
+export interface ReplayUserInteractionsResultShort {
+  length: "short";
+  reason: "max events" | "max duration";
+}
```
Que3216, line 43: Nice

packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts
```diff
@@ -1,3 +1,43 @@
...
+  acceleratePlayback: boolean;
+  virtualTime: VirtualTimeOptions;
+  maxDurationMs?: number;
+  maxEventCount?: number;
+  onTimelineEvent: OnReplayTimelineEventFn;
+  logLevel: LogLevelDesc;
+}
+
+/** Result of replaying user interactions */
+export type ReplayUserInteractionsResult =
```
Que3216, line 19: This type should probably sit in `replay/bundle-to-sdk` folder, and then be imported here. This is because e.g. it's non-breaking to add a new required field to ReplayUserInteractionsResultFull, but breaking to add a new required field to ReplayUserInteractionsOptions; it's non-breaking to delete a field from ReplayUserInteractionsOptions but breaking to delete one from ReplayUserInteractionsResultFull etc. `bundle-to-sdk` would then have a different README.md in it.

(or even better we move ReplayUserInteractionsResult to `replay/bundle-to-sdk`, and OnReplayTimelineEventFn to `sdk-bundles-api/src/functions.ts`, or something and `functions.ts` imports from both bundle-to-sdk and sdk-to-bundle, thereby reflecting the fact that ReplayUserInteractionsFn encodes messages that travel in both directions).

Que3216 (approved):
This looks great! I think having these types will speed us up. Added one comment, but otherwise looks good.

packages/sdk-bundles-api/package.json
```diff
@@ -17,6 +17,10 @@
     "lint:commit": "eslint --cache $(git diff --relative --name-only --diff-filter=ACMRTUXB master | grep  -E \"(.js$|.ts$|.tsx$)\")",
     "lint:fix": "eslint src --ext=ts,tsx,js --cache --fix"
   },
+  "devDependencies": {
```
Que3216, line 20: Are these devDependencies or peerDependencies? (can't remember details of how peer dependencies work / are inherited; but if they do work as desired for this then it could be quite nice having them as peerDependencies, because it'd enforce both main repo & sdk repo to use same version of puppeteer -- using separate versions of puppeteer may break things)

leafty: Right :+1:

packages/sdk-bundles-api/src/replay/bundle-to-sdk/README.md
```diff
@@ -0,0 +1,14 @@
+### APIs used for messages sent from the replay and record bundles to the SDK
```
leafty, line 1: @Que3216 Can you double-check what this says about updates?

Que3216 (approved):

PR #131, opened 2022-10-24
Deduplicate replay interfaces

Que3216:
Cleans up our replay options interfaces (now that I added some extra options here it was getting a bit verbose).

### 1. Grouping options together

image

For example ReplayExecutionOptions and ScreenshottingOptions. This hopefully makes the code easier to follow, because it reduces the number of options and so makes it easier for readers to see which options are different vs which options are the same between the different interfaces. It should hopefully make it less verbose to pass things around (e.g. often the ReplayExecutionOptions were just getting passed from one method to the next without getting changed).

### Centralizing defaults at the command level

We try to centralize all defaults at the command level, since then yargs prints out the default value to the user on the help screen, which can be useful for the user. This also means that we can keep the descriptions, defaults etc. for the command in sync; and it's very explicit/clear whenever they do differ from one command vs another.

## Notes

 - The intent here is that bundling into the ReplayExecutionOptions and ScreenshottingOptions is done as soon as possible, right at the start of the command handler. And then the rest of the program mainly just passes around the bundled types.
 - I've tried to use `| undefined` when we want to force the caller to think about what value they want to pass for the args, and only use `?` forgetting the value is unlikely to cause issues & always supplying an undefined value would be annoying/verbose.
 - I haven't updated the other commands or all the other options to match this pattern. Can do so in future PRs, or whenever we next edit those commands.
 - Tried to ensure the types can only represent valid combinations (e.g. see ScreenshottingOptions and ReplayTarget), so that it's easier for readers to understand the options & how they relate; and harder to introduce bugs.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -34,15 +34,13 @@
     "express": "^4.18.1",
     "find-free-port": "^2.0.0",
     "inquirer": "^8.2.4",
-    "lodash": "^4.17.21",
     "luxon": "^2.4.0",
     "pixelmatch": "^5.3.0",
     "pngjs": "^6.0.0",
     "yargs": "^17.5.1"
   },
   "devDependencies": {
-    "@types/express": "^4.17.14",
-    "@types/lodash": "^4.14.186"
+    "@types/express": "^4.17.14"
   },
   "peerDependencies": {
     "@alwaysmeticulous/record": "^2.2.1",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -34,13 +34,15 @@
     "express": "^4.18.1",
     "find-free-port": "^2.0.0",
     "inquirer": "^8.2.4",
+    "lodash": "^4.17.21",
     "luxon": "^2.4.0",
     "pixelmatch": "^5.3.0",
     "pngjs": "^6.0.0",
     "yargs": "^17.5.1"
   },
   "devDependencies": {
-    "@types/express": "^4.17.14"
+    "@types/express": "^4.17.14",
+    "@types/lodash": "^4.14.186"
   },
   "peerDependencies": {
     "@alwaysmeticulous/record": "^1.7.6",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -45,9 +45,9 @@
     "@types/lodash": "^4.14.186"
   },
   "peerDependencies": {
-    "@alwaysmeticulous/record": "^1.7.6",
-    "@alwaysmeticulous/replay-debugger": "^1.7.6",
-    "@alwaysmeticulous/replayer": "^1.7.6"
+    "@alwaysmeticulous/record": "^2.2.1",
+    "@alwaysmeticulous/replay-debugger": "^2.2.1",
+    "@alwaysmeticulous/replayer": "^2.2.1"
   },
   "peerDependenciesMeta": {
     "@alwaysmeticulous/record": {
--- a/packages/cli/src/command-utils/command-utils.ts
+++ b/packages/cli/src/command-utils/command-utils.ts
@@ -6,7 +6,7 @@ type UndefinedToNullOrUndefined<T extends object> = {
  * yargs normally passes through undefineds, but passes through nulls for
  * numbers which fail to parse. We convert everything to undefined to standardize.
  */
-export const convertNullsToUndefineds = <T extends object>(
+export const handleNulls = <T extends object>(
   handler: (args: T) => Promise<void>
 ): ((args: UndefinedToNullOrUndefined<T>) => Promise<void>) => {
   return (args: UndefinedToNullOrUndefined<T>) => {
--- a/packages/cli/src/command-utils/command-utils.ts
+++ b/packages/cli/src/command-utils/command-utils.ts
@@ -1,0 +1,23 @@
+type UndefinedToNullOrUndefined<T extends object> = {
+  [K in keyof T]: undefined extends T[K] ? T[K] | null : T[K];
+};
+
+/**
+ * yargs normally passes through undefineds, but passes through nulls for
+ * numbers which fail to parse. We convert everything to undefined to standardize.
+ */
+export const convertNullsToUndefineds = <T extends object>(
+  handler: (args: T) => Promise<void>
+): ((args: UndefinedToNullOrUndefined<T>) => Promise<void>) => {
+  return (args: UndefinedToNullOrUndefined<T>) => {
+    const keys = Object.keys(args) as Array<keyof T>;
+    const newArgs = keys.reduce(
+      (partialArgs: Partial<T>, key: keyof T): Partial<T> => ({
+        ...partialArgs,
+        [key]: args[key] === null ? undefined : args[key],
+      }),
+      {}
+    ) as T;
+    return handler(newArgs);
+  };
+};
--- a/packages/cli/src/command-utils/common-options.ts
+++ b/packages/cli/src/command-utils/common-options.ts
@@ -1,0 +1,65 @@
+export const DEFAULT_MISMATCH_THRESHOLD = 0.01;
+
+// Matches https://github.com/mapbox/pixelmatch/blob/master/index.js#L6
+export const DEFAULT_DIFF_PIXEL_THRESHOLD = 0.1;
+
+export const SCREENSHOT_DIFF_OPTIONS = {
+  diffThreshold: {
+    number: true,
+    description:
+      "Acceptable maximum proportion of changed pixels, between 0 and 1. If this proportion is exceeded then the test will fail.",
+    default: DEFAULT_MISMATCH_THRESHOLD,
+  },
+  diffPixelThreshold: {
+    number: true,
+    description:
+      "A number between 0 and 1. Color/brightness differences in individual pixels will be ignored if the difference is less than this threshold. A value of 1.0 would accept any difference in color, while a value of 0.0 would accept no difference in color.",
+    default: DEFAULT_DIFF_PIXEL_THRESHOLD,
+  },
+};
+
+export const PRIMARY_COMMON_REPLAY_OPTIONS = {
+  apiToken: {
+    string: true,
+  },
+  commitSha: {
+    string: true,
+  },
+};
+
+export const SECONDARY_COMMON_REPLAY_OPTIONS = {
+  headless: {
+    boolean: true,
+    description: "Start browser in headless mode",
+  },
+  devTools: {
+    boolean: true,
+    description: "Open Chrome Dev Tools",
+  },
+  bypassCSP: {
+    boolean: true,
+    description: "Enables bypass CSP in the browser",
+  },
+  padTime: {
+    boolean: true,
+    description:
+      "Pad replay time according to recording duration. Please note this option will be ignored if running with the '--accelerate' option.",
+    default: true,
+  },
+  shiftTime: {
+    boolean: true,
+    description: "Shift time during simulation to be set as the recording time",
+    default: true,
+  },
+  networkStubbing: {
+    boolean: true,
+    description: "Stub network requests during replay",
+    default: true,
+  },
+  accelerate: {
+    boolean: true,
+    description:
+      "Fast forward through any pauses to replay as fast as possible. Warning: this option is experimental and may be deprecated",
+    default: false,
+  },
+};
--- a/packages/cli/src/command-utils/common-options.ts
+++ b/packages/cli/src/command-utils/common-options.ts
@@ -1,20 +1,13 @@
-export const DEFAULT_MISMATCH_THRESHOLD = 0.01;
-
-// Matches https://github.com/mapbox/pixelmatch/blob/master/index.js#L6
-export const DEFAULT_DIFF_PIXEL_THRESHOLD = 0.1;
-
 export const SCREENSHOT_DIFF_OPTIONS = {
   diffThreshold: {
     number: true,
     description:
       "Acceptable maximum proportion of changed pixels, between 0 and 1. If this proportion is exceeded then the test will fail.",
-    default: DEFAULT_MISMATCH_THRESHOLD,
   },
   diffPixelThreshold: {
     number: true,
     description:
       "A number between 0 and 1. Color/brightness differences in individual pixels will be ignored if the difference is less than this threshold. A value of 1.0 would accept any difference in color, while a value of 0.0 would accept no difference in color.",
-    default: DEFAULT_DIFF_PIXEL_THRESHOLD,
   },
 };
 
--- a/packages/cli/src/command-utils/common-options.ts
+++ b/packages/cli/src/command-utils/common-options.ts
@@ -1,37 +1,36 @@
-export const SCREENSHOT_DIFF_OPTIONS = {
-  diffThreshold: {
-    number: true,
-    description:
-      "Acceptable maximum proportion of changed pixels, between 0 and 1. If this proportion is exceeded then the test will fail.",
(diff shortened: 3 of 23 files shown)


leafty:
Looks promising :smile:

packages/cli/src/commands/record/record.command.ts
```diff
@@ -18,15 +18,15 @@ import { getCommitSha } from "../../utils/commit-sha.utils";
...
-  trace?: boolean | null | undefined;
+  apiToken: string | undefined;
+  commitSha: string | undefined;
+  devTools: boolean | undefined;
+  bypassCSP: boolean | undefined;
+  width: number | undefined | null;
+  height: number | undefined | null;
+  uploadIntervalMs: number | undefined | null;
+  incognito: boolean | undefined;
+  trace: boolean | undefined;
```
Que3216, line 29: yargs will only provide nulls in the case of numbers that don't pass e.g. `--width "I'm a string"`. For strings and booleans it'll only ever provide string/boolean or undefined (specifically absent key). So simplified our typings a bit to be more accurate.

Note that to be truly accurate (and so ensure correct usage of things like `{...spread}` operations), we'd want these as optional params:

```
apiToken?: string;
```

However we have many commands that call each other, and I want to make sure we don't accidentally forget to wire through new flags etc. Using `| undefined` instead of `?` is the way to force that.

packages/cli/src/commands/record/record.command.ts
```diff
@@ -17,16 +17,19 @@ import { fetchAsset } from "../../local-data/replay-assets";
...
```
Que3216, line 32: yargs will only provide nulls in the case of numbers that don't pass e.g. `--width "I'm a string"`. For strings and booleans it'll only ever provide string/boolean or undefined (specifically absent key). So simplified our typings a bit to be more accurate.

Note that to be truly accurate (and so ensure correct usage of things like `{...spread}` operations), we'd want these as optional params:

```
apiToken?: string;
```

However in the cases where we have commands that call each other, and I want to make sure we don't accidentally forget to wire through new flags / the compiler forces us to think about all the cases when we add a new flag etc. Using `| undefined` instead of `?` is the way to force that.

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -312,23 +284,134 @@ export const replayCommandHandler: (
...
+  if (replayTarget.type === "url") {
+    return { appUrl: replayTarget.appUrl };
+  }
+  if (replayTarget.type === "original-recorded-url") {
+    return {};
+  }
+  return assertNever(replayTarget);
+};
+
+const assertNever = (value: never): never => {
```
Que3216, line 310: I don't think we use this pattern anywhere yet, but this is a common ts trick used to get the compiler to tell you if you've forgotten to handle a discriminated type union case anywhere (useful in the case we added in future e.g. a fourth type of replay target; the compiler would force us to update all the right places) .e.g https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript

Side note: with strictNullChecks enabled and a function that has a required return type you can just omit the last return statement, and get the same effect. However we have a compiler setting or eslint rule that forces functions that return a value to have a final return statemen -- but assertNever works.

Que3216: If we find ourselves using this a lot we can put it in a common package.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -36,24 +36,23 @@ export const diffScreenshots: (options: {
...
   headScreenshot,
-  threshold: threshold_,
-  pixelThreshold,
+  diffOptions,
   exitOnMismatch,
 }) => {
+  const { diffThreshold, diffPixelThreshold } = diffOptions;
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
 
-  const threshold = threshold_ || DEFAULT_MISMATCH_THRESHOLD;
```
Que3216, line 54: Previously there was a bug here in that it was impossible to set the threshold to 0 (this is a danger of using `||` by default instead of `??`)

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -36,24 +36,23 @@ export const diffScreenshots: (options: {
...
-  const pixelmatchOptions: CompareImageOptions["pixelmatchOptions"] | null =
-    pixelThreshold ? { threshold: pixelThreshold } : null;
```
Que3216, line 56: Similar to above previously there was a bug here in that it was impossible to override pixelThreshold to 0 - setting it to 0 would set it to 0.1 (this is a danger of using falsey/truthy checks by default instead of `!= null` checks, and explicit `... != null && ... !== 0` checks when you do want to handle 0 and null case at same time)

packages/cli/src/config/config.types.ts
```diff
@@ -1,21 +1,23 @@
-export interface ReplayOptions {
-  screenshotSelector?: string;
-  diffThreshold?: number;
-  diffPixelThreshold?: number;
-  cookies?: Record<string, any>[];
```
Que3216, line 5: `cookies` were never used (I searched the codebase)

packages/cli/package.json
```diff
@@ -34,18 +34,20 @@
...
+    "@types/express": "^4.17.14",
+    "@types/lodash": "^4.14.186"
   },
   "peerDependencies": {
-    "@alwaysmeticulous/record": "^1.7.6",
-    "@alwaysmeticulous/replay-debugger": "^1.7.6",
-    "@alwaysmeticulous/replayer": "^1.7.6"
+    "@alwaysmeticulous/record": "^2.2.1",
+    "@alwaysmeticulous/replay-debugger": "^2.2.1",
+    "@alwaysmeticulous/replayer": "^2.2.1"
```
Que3216, line 50: I bumped the dependencies here since I made breaking changes in the API between these packages and the CLI, and so want to ensure that when people bump the CLI they are forced to bump the other packages too. These versions haven't actually been released yet, I will cut a release after this merges.

Side note: presumably our customers are getting peer dependency errors now if they're on replayer 2.* and cli 2.*, since the dependency was set to `^1.7.6`. I guess these show as warns not errors?

packages/replayer/src/replayer.ts
```diff
@@ -31,27 +31,26 @@ export const replayEvents: ReplayEventsFn = async (options) => {
...
     options: {
       appUrl,
       outputDir,
-      headless,
-      devTools,
       dependencies,
-      padTime,
-      shiftTime,
-      networkStubbing,
+      ...replayExecutionOptions,
```
Que3216, line 53: Note that previously this was missing some of the options we added recently, like acceleration. Going forward that shouldn't be a risk since they'll be added automatically (which in this case is what we presumably want - just capture all the options).

packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
```diff
@@ -99,55 +118,34 @@ const handler: (options: Options) => Promise<void> = async ({
...
-        headless,
-        devTools,
-        bypassCSP,
-        diffThreshold,
-        diffPixelThreshold,
-        padTime,
-        shiftTime,
-        networkStubbing,
-        parallelTasks,
+        parallelTasks: parallelTasks ?? undefined,
```
Que3216, line 127: The fact that numbers can come through as null is an implementation detail of the fact we use yargs, so I tried to isolate this as early as possible, and not let the nulls propagate too far through our codebase (just use undefineds instead). Note that optional params, absent keys etc. translate into undefined, so if we want to only use one of null vs undefined, it's normally easier to just use undefined (this is the reason we standardised on undefined at Palantir, and I think also [at Microsoft](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined)).

packages/replayer/src/screenshot.utils.ts
```diff
@@ -7,13 +7,13 @@ import { prepareScreenshotsDir } from "./output.utils";
 export interface TakeScreenshotOptions {
   page: Page;
   outputDir: string;
-  screenshotSelector: string;
+  screenshotSelector: string | undefined;
```
Que3216, line 10: I changed the cases where we were previously converting to empty string, to instead passing undefined through.

Converting to empty string makes sense for raw javascript where you don't have strict null checks, and so want to reduce the chance of a null pointer exception, and so I've often seen it in JS codebases. But for typescript the tradeoffs are different. When its typed as just string the reader has no hint that the screenshot selector could be absent (represented as ""), and so may not think to handle this case. However when it's `string | undefined` they'll know that it may be absent, and will normally be forced by typescript to handle the absent case. I've found  therefore that the explicit | undefined typings can reduce bugs and aids understanding for other readers.

packages/replayer/src/replayer.ts
```diff
@@ -69,8 +68,8 @@ export const replayEvents: ReplayEventsFn = async (options) => {
         // including the respective Preflgiht CORS requests which are not handled by the network stubbing layer.
         "--disable-web-security",
       ],
-      headless: headless || false,
-      devtools: devTools || false,
```
Que3216, line 73: We've centralized all defaults to the command level.

packages/cli/src/deflake-tests/deflake-tests.handler.ts
```diff
@@ -38,27 +71,74 @@ const handleReplay: (options: {
...
+  // Options specified in the test case override those passed as CLI flags
+  // (CLI flags set the defaults)
+  return defaults(
+    {},
+    {
+      moveBeforeClick: overridesFromTestCase.moveBeforeClick,
+      simulationIdForAssets: overridesFromTestCase.simulationIdForAssets,
+    },
+    executionOptionsFromCliFlags
+  );
```
leafty, line 120: We can use object destructuring instead of lodash here.

Que3216: Object destructing, or Object.assign, won't handle undefined values correctly. We could write our own defaults function using Object.keys, but lodash/defaults should be pretty small.

packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
leafty: > STOP READING IMMEDIATELY
> THIS PAGE PROBABLY DOES NOT PERTAIN TO YOU.
> 
> These are Coding Guidelines for Contributors to TypeScript. This is NOT a prescriptive guideline for the TypeScript community. These guidelines are meant for contributors to the TypeScript project's codebase. We have chosen many of them for team consistency. Feel free to adopt them for your own team.
> 
> AGAIN: This is NOT a prescriptive guideline for the TypeScript community

I previously tried `undefined` everywhere but it ended up reading worse in many circumstances

Que3216: Yep those are TS's internal guidelines, but I think they use most of them more widely across Microsoft. I've found them useful, because although the tradeoffs for TypeScript vs other codebases can be different, the TS team has thought about them carefully.

> I previously tried undefined everywhere but it ended up reading worse in many circumstances

Do you have an example? Is this re. using undefined vs null? Are you proposing using null instead of undefined, or using empty string instead of null or undefined?

Can just chat when we catch up if that's easier.

packages/cli/src/deflake-tests/deflake-tests.handler.ts
Que3216, line 120:
```ts
return {
       ... executionOptionsFromCliFlags,
      moveBeforeClick: overridesFromTestCase.moveBeforeClick ?? executionOptionsFromCliFlags.moveBeforeClick,
      simulationIdForAssets: overridesFromTestCase.simulationIdForAssets ?? executionOptionsFromCliFlags.simulationIdForAssets,
    }
```

packages/cli/src/commands/record/record.command.ts
```diff
@@ -17,16 +17,19 @@ import { fetchAsset } from "../../local-data/replay-assets";
...
 export interface RecordCommandHandlerOptions {
-  apiToken?: string | null | undefined;
-  commitSha?: string | null | undefined;
-  devTools?: boolean | null | undefined;
-  bypassCSP?: boolean | null | undefined;
-  width?: number | null | undefined;
-  height?: number | null | undefined;
-  uploadIntervalMs?: number | null | undefined;
-  incognito?: boolean | null | undefined;
-  trace?: boolean | null | undefined;
```
Que3216, line 29: TODO:

```ts
type UndefinedToNullOrUndefined<T> = { [K in keyof T]: undefined extends T[K] ? T[K] | null : T[K] };

const convertNullsToUndefineds = <T>(record: UndefinedToNullOrUndefined<T>): T => {
     return Object.keys(x).reduce(
           (partialRecord, key) => ({ ...partialRecord, [key]: x[key] === null ? undefined : x[key] }),
     {});
}
```

packages/cli/src/command-utils/common-options.ts
```diff
@@ -0,0 +1,92 @@
...
+    number: true,
+    description:
+      "Acceptable maximum proportion of changed pixels, between 0 and 1. If this proportion is exceeded then the test will fail.",
+    default: 0.01,
+  },
+  diffPixelThreshold: {
+    number: true,
+    description:
+      "A number between 0 and 1. Color/brightness differences in individual pixels will be ignored if the difference is less than this threshold. A value of 1.0 would accept any difference in color, while a value of 0.0 would accept no difference in color.",
+    default: 0.1, // matches https://github.com/mapbox/pixelmatch/blob/master/index.js#L6
```
leafty, line 72: Note: I think we were using `0.01` as the default before this PR.

packages/cli/src/commands/record/record.command.ts
```diff
@@ -13,20 +13,24 @@ import {
...
   postSessionIdNotification,
 } from "../../api/session.api";
+import { convertNullsToUndefineds } from "../../command-utils/command-utils";
 import { fetchAsset } from "../../local-data/replay-assets";
 import { getCommitSha } from "../../utils/commit-sha.utils";
 import { wrapHandler } from "../../utils/sentry.utils";
 
+// Note: We use `| undefined` instead of `?` here, despite yargs passing through absent keys for
+// unspecified options. We do this because recordCommandHandler is re-used by create-test
+// and we want to enforce that create-test passes through the full set of options, and does not forget any.
```
leafty, line 23: I think this comment can be removed. The interface itself is clear I think.

packages/cli/src/command-utils/command-utils.ts
```diff
@@ -0,0 +1,23 @@
+type UndefinedToNullOrUndefined<T extends object> = {
+  [K in keyof T]: undefined extends T[K] ? T[K] | null : T[K];
+};
+
+/**
+ * yargs normally passes through undefineds, but passes through nulls for
+ * numbers which fail to parse. We convert everything to undefined to standardize.
+ */
+export const convertNullsToUndefineds = <T extends object>(
```
leafty, line 9:
```suggestion
export const handleNulls = <T extends object>(
```

packages/cli/src/command-utils/common-options.ts
Que3216: Ah ok, I'll double check

Que3216: Ah yes you're right. We apply a default of 0.01 in compareImages in diff.utils that I missed. I'll fix this.

PR #132, opened 2022-10-24
add support for diffing multiple simulation screenshots

alexivanov:
### Changes

We're adding the ability to take multiple screenshots during simulations.

This PR extends the existing diffing utilities to handle comparing  simulations with multiple screenshots.

In cases where we have multiple screenshots diffing can "fail" if any of the following is true:
 * for any screenshot pair, the pixel mismatch is above the set threshold
 * overall pixel mismatch across all screenshot pairs is above the set threshold
 * the head simulation is missing a screenshot which is present in the base simulation-- currently we assume the base simulation screenshots are a subset of the the head simulation.

The failing criteria can change in the future but this preserves the current behaviour (as `final-state` screenshots is a special case of the multiple screenshot simulations)

### DIFF
--- a/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
+++ b/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
@@ -5,7 +5,7 @@ import { PNG } from "pngjs";
 import { CommandModule } from "yargs";
 import { createClient } from "../../api/client";
 import { getDiffUrl, postScreenshotDiffStats } from "../../api/replay.api";
-import { CompareImageOptions, compareImages } from "../../image/diff.utils";
+import { compareImages } from "../../image/diff.utils";
 import {
   getOrFetchReplay,
   getOrFetchReplayArchive,
@@ -52,14 +52,12 @@ export const diffScreenshots: (options: {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
 
   const threshold = threshold_ || DEFAULT_MISMATCH_THRESHOLD;
-  const pixelmatchOptions: CompareImageOptions["pixelmatchOptions"] | null =
-    pixelThreshold ? { threshold: pixelThreshold } : null;
 
   try {
     const { mismatchPixels, mismatchFraction, diff } = compareImages({
       base: baseScreenshot,
       head: headScreenshot,
-      ...(pixelmatchOptions ? pixelmatchOptions : {}),
+      pixelThreshold: pixelThreshold ?? null,
     });
 
     logger.debug({ mismatchPixels, mismatchFraction });
--- a/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
+++ b/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
@@ -242,7 +242,7 @@ const logComparisonResultMessage: (
   message: string,
   outcome: ComparisonOutcome
 ) => void = (logger, message, outcome) => {
-  logger.info(`${message} => ${outcome === "fail" ? "FAIL!" : "PASS"}`);
+  logger.info(`${message} => ${outcome === "pass" ? "PASS" : "FAIL!"}`);
 };
 
 interface Options {
--- a/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
+++ b/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
@@ -1,18 +1,21 @@
 import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
 import { AxiosInstance } from "axios";
-import log from "loglevel";
-import { PNG } from "pngjs";
+import log, { Logger } from "loglevel";
 import { CommandModule } from "yargs";
 import { createClient } from "../../api/client";
 import { getDiffUrl, postScreenshotDiffStats } from "../../api/replay.api";
-import { compareImages } from "../../image/diff.utils";
+import { CompareImageResult, compareImages } from "../../image/diff.utils";
 import {
   getOrFetchReplay,
   getOrFetchReplayArchive,
-  readReplayScreenshot,
+  getReplayDir,
+  getScreenshotFiles,
+  getScreenshotsDir,
 } from "../../local-data/replays";
 import { writeScreenshotDiff } from "../../local-data/screenshot-diffs";
 import { wrapHandler } from "../../utils/sentry.utils";
+import { readPng } from "../../image/io.utils";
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -23,79 +26,198 @@ export class DiffError extends Error {
...
-  headScreenshot,
+  baseScreenshotsDir,
+  headScreenshotsDir,
   threshold: threshold_,
   pixelThreshold,
   exitOnMismatch,
 }) => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
 
   const threshold = threshold_ || DEFAULT_MISMATCH_THRESHOLD;
```
leafty, line 67: I think this is a mistake from me:

```suggestion
  const threshold = threshold_ ?? DEFAULT_MISMATCH_THRESHOLD;
```

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -23,79 +26,198 @@ export class DiffError extends Error {
...
+          baseReplayId,
+          headReplayId,
+          threshold,
+        }
+      );
+    }
+
+    const totalMismatchedFraction = totalMismatchPixels / totalComparedPixels;
+    const overallMismatchOutcome: ComparisonOutcome =
+      totalMismatchedFraction > threshold ? "fail" : "pass";
```
leafty, line 199: I think doing the overall check is not needed (it's impossible to fail).

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -111,6 +233,16 @@ export const diffScreenshots: (options: {
...
+
+  return comparisonResults;
+};
+
+const logComparisonResultMessage: (
+  logger: Logger,
+  message: string,
+  outcome: ComparisonOutcome
+) => void = (logger, message, outcome) => {
+  logger.info(`${message} => ${outcome === "fail" ? "FAIL!" : "PASS"}`);
```
leafty, line 245: nitpick: more foolproof if there are more outcome types later.

```suggestion
  logger.info(`${message} => ${outcome === "pass" ? "PASS" : "FAIL!"}`);
```

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
alexivanov: ah yes, good point. AVG > threshold will be caught by the above individual screenshot checks https://github.com/alwaysmeticulous/meticulous-sdk/pull/132/commits/4ac49190537a6b06906cb85e0a2cf57b08e0b9c2

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -23,79 +26,172 @@ export class DiffError extends Error {
...
+      comparisonResults.push({
+        baseScreenshotFile,
+        headScreenshotFile,
+        comparisonResult,
+        outcome:
+          comparisonResult.mismatchFraction > threshold ? "fail" : "pass",
+      });
+    }
 
     await postScreenshotDiffStats(client, {
```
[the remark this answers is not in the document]
leafty: I think we can do that by updating the [TestRun](https://github.com/alwaysmeticulous/meticulous/blob/master/packages/database-entities/src/test-run/test-run.entity.ts) entity and processing on the backend, yes :+1:

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -23,79 +26,172 @@ export class DiffError extends Error {
...
+      const headScreenshot = await readPng(headScreenshotFile);
+
+      const comparisonResult = compareImages({
+        base: baseScreenshot,
+        head: headScreenshot,
+        pixelThreshold: pixelThreshold ?? null,
+      });
+
+      totalMismatchPixels += comparisonResult.mismatchPixels;
+      totalComparedPixels += baseScreenshot.width * baseScreenshot.height;
```
leafty, line 122: Reported not used

leafty (approved):
LGTM :clap:

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -23,79 +26,172 @@ export class DiffError extends Error {
...
+  const threshold = threshold_ ?? DEFAULT_MISMATCH_THRESHOLD;
+
+  const baseReplayScreenshots = await getScreenshotFiles(baseScreenshotsDir);
+  const headReplayScreenshots = await getScreenshotFiles(headScreenshotsDir);
+
+  // Assume base replay screenshots are always a subset of the head replay screenshots.
+  // We report any missing base replay screenshots for visibility but don't count it as a difference.
+  const missingHeadImages = new Set(
+    [...baseReplayScreenshots].filter(
+      (file_) => !headReplayScreenshots.includes(file_)
```
leafty, line 76: I forgot this one:

```suggestion
      (file) => !headReplayScreenshots.includes(file)
```

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -1,18 +1,21 @@
...
   getOrFetchReplayArchive,
-  readReplayScreenshot,
+  getReplayDir,
+  getScreenshotFiles,
+  getScreenshotsDir,
 } from "../../local-data/replays";
 import { writeScreenshotDiff } from "../../local-data/screenshot-diffs";
 import { wrapHandler } from "../../utils/sentry.utils";
+import { readPng } from "../../image/io.utils";
+import { basename, join } from "path";
```
leafty, line 18: Nitpick: run "optimize imports"

leafty (commented):
Forgot these comments

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
alexivanov: My IDE thinks imports are ordered correctly here 🤔 wonder if should standardise import ordering with eslint / prettier?

alexivanov: https://github.com/alwaysmeticulous/meticulous-sdk/pull/132/commits/eff308933321daf8e9a050292f84bddfc83d3f3c

PR #135, opened 2022-10-26
Replay: add support for max duration and max events

leafty:
[SIM-307](https://toil.kitemaker.co/qpgVHi-Meticulous/pPGwrM-Engineering/items/307)

### DIFF
--- a/packages/cli/src/commands/replay/replay.command.ts
+++ b/packages/cli/src/commands/replay/replay.command.ts
@@ -61,6 +61,8 @@ export interface ReplayCommandHandlerOptions {
   cookies?: Record<string, any>[];
   cookiesFile?: string | null | undefined;
   accelerate: boolean;
+  maxDurationMs?: number | null | undefined;
+  maxEventCount?: number | null | undefined;
 }
 
 export const replayCommandHandler: (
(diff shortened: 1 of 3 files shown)


packages/replayer/src/replayer.ts
```diff
@@ -139,6 +141,12 @@ export const replayEvents: ReplayEventsFn = async (options) => {
     startUrl,
   });
 
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const userInteractionsModule = require(dependencies.nodeUserInteractions
+    .location);
+  const replayUserInteractions =
+    userInteractionsModule.replayUserInteractions as ReplayUserInteractionsFn;
```
leafty, line 148: Note: I think it would be even possible to type the whole bundle as a module.

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -418,6 +424,14 @@ export const replay: CommandModule<unknown, ReplayCommandHandlerOptions> = {
...
       default: false,
     },
+    maxDurationMs: {
+      number: true,
+      description: "Max duration",
+    },
+    maxEventCount: {
+      number: true,
+      description: "Max event count",
+    },
```
leafty, line 434: TODO: add full text descriptions

Que3216 (approved):

PR #137, opened 2022-10-28
Add types to control the storyboard generation feature

leafty:


packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -57,6 +57,7 @@ export interface ReplayOptions extends AdditionalReplayOptions {
   replayTarget: ReplayTarget;
   executionOptions: ReplayExecutionOptions;
   screenshottingOptions: ScreenshotAssertionsOptions;
+  storyboard: boolean;
```
Que3216, line 60: Would it be possible to put this inside `ScreenshotAssertionsEnabledOptions` instead? It should make it easier to wire through (will get auto-wired through to the right places), and will make it clear that passing this option is invalid if you pass `screenshottingOptions = { enabled: false }`.

leafty: Done

Que3216 (approved):
Looks great!

PR #139, opened 2022-10-28
Rename accelerate to skipPauses

Que3216:


packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts
```diff
@@ -8,8 +8,7 @@ export interface ReplayUserInteractionsOptions {
   page: Page;
   sessionData: unknown;
   moveBeforeClick: boolean;
-  acceleratePlayback: boolean;
```
Que3216, line 11: This is obsolete now

packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts
```diff
@@ -8,8 +8,7 @@ export interface ReplayUserInteractionsOptions {
...
-  virtualTime: VirtualTimeOptions;
+  virtualTime?: VirtualTimeOptions;
```
Que3216, line 11: Since this was an added field it needs to be optional to ensure back compat when consuming these types from the bundle side.

Gabriel-H (approved):

PR #140, opened 2022-10-28
fix dist imports

Gabriel-H:
Follow-up to https://github.com/alwaysmeticulous/meticulous-sdk/pull/131/files, we still have two versions of `StoryboardOptions` floating about:

1) https://github.com/alwaysmeticulous/meticulous-sdk/blob/a7333c8e681abd136ca4e1526a24a4f2ae54451a/packages/common/src/types/replay.types.ts#L79

2) https://github.com/alwaysmeticulous/meticulous-sdk/blob/a7333c8e681abd136ca4e1526a24a4f2ae54451a/packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts#L30

Will fix that in follow-up PR.

### DIFF
--- a/packages/cli/src/deflake-tests/deflake-tests.handler.ts
+++ b/packages/cli/src/deflake-tests/deflake-tests.handler.ts
@@ -1,16 +1,15 @@
-import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
-import { ReplayExecutionOptions, ReplayTarget } from "@alwaysmeticulous/common";
+import { METICULOUS_LOGGER_NAME, ReplayExecutionOptions, ReplayTarget } from "@alwaysmeticulous/common";
 import log from "loglevel";
 import {
   ScreenshotAssertionsEnabledOptions,
-  ScreenshotDiffOptions,
+  ScreenshotDiffOptions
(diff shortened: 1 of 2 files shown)


Que3216:
Oh I think the two storyboard options are intended, though we should rename one. One includes the screenshotsDir, to be passed into the replay function. Maybe we call that StoryboardConfig or something.

packages/cli/src/deflake-tests/deflake-tests.handler.ts
```diff
@@ -1,8 +1,5 @@
 import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
-import {
-  ReplayExecutionOptions,
-  ReplayTarget,
-} from "@alwaysmeticulous/common/dist/types/replay.types";
+import { ReplayExecutionOptions, ReplayTarget } from "@alwaysmeticulous/common";
```
Que3216, line 2: You'll need to combine this with import above for lint to not fail. You can use Command + Shift + P -> Optimize Imports in vscode, or just merge in develop and hit save and the auto-fix should solve it.

Que3216 (approved):

PR #144, opened 2022-10-31
Add ability to hold browser open

Que3216:
Very useful for debugging.

Putting this as a env var, rather than a param, since it's really designed for our internal use.


Gabriel-H (approved):
Used to have `--hold` but looks like that got deleted

PR #145, opened 2022-10-31
Make yargs usage typesafe

Que3216:


packages/cli/src/commands/update-tests/update-tests.command.ts
```diff
@@ -109,6 +109,7 @@ export const updateTests: CommandModule<unknown, Options> = {
       description: "Test run id to fix",
     },
     accept: {
+      string: true,
```
Que3216, line 112: It's an array of strings (I tested and this works as intended)

packages/cli/src/commands/create-test/create-test.command.ts
```diff
@@ -20,7 +19,20 @@ import {
...
+      | "simulationIdForAssets"
+      | "maxDurationMs"
+      | "maxEventCount"
+      | "storyboard"
+      | "diffThreshold"
+      | "diffPixelThreshold"
+      | "sessionId"
+      | "save"
+      | "baseSimulationId"
+    > {}
```
Que3216, line 35: We have to Omit stuff now, since it forces us to have correct typings. Before we were falsing implying in our types that we had valid numerical values for things like diffThreshold, when they were actually undefined.

packages/cli/src/commands/create-test/create-test.command.ts
```diff
@@ -204,10 +217,7 @@ export const createTest: CommandModule<unknown, Options> = {
       description:
         "Query selector to screenshot a specific DOM element instead of the whole page",
     },
-    moveBeforeClick: {
-      boolean: true,
-      description: "Simulate mouse movement before clicking",
-    },
+    moveBeforeClick: OPTIONS.moveBeforeClick,
```
Que3216, line 220: Fixing the types caught a bug that we weren't providing the default value for moveBeforeClick.

Gabriel-H (approved):
LGTM

PR #149, opened 2022-11-03
Introduce disable remote fonts as option

Gabriel-H:
This change also changes the replay-debugger command to use `COMMON_REPLAY_OPTIONS`.

### DIFF
--- a/packages/cli/src/commands/debug-replay/debug-replay.command.ts
+++ b/packages/cli/src/commands/debug-replay/debug-replay.command.ts
@@ -117,6 +117,9 @@ export const debugReplay = buildCommand("debug-simulation")
       string: true,
       description: "Path to cookies to inject before replay",
     },
-    ...COMMON_REPLAY_OPTIONS
+    devTools: COMMON_REPLAY_OPTIONS.devTools,
+    shiftTime: COMMON_REPLAY_OPTIONS.shiftTime,
+    networkStubbing: COMMON_REPLAY_OPTIONS.networkStubbing,
+    disableRemoteFonts: COMMON_REPLAY_OPTIONS.disableRemoteFonts,
   })
   .handler(handler);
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/debug-replay/debug-replay.command.ts
```diff
@@ -128,5 +117,6 @@ export const debugReplay = buildCommand("debug-simulation")
       string: true,
       description: "Path to cookies to inject before replay",
     },
+    ...COMMON_REPLAY_OPTIONS
```
Que3216, line 120: This adds a bunch of options which aren't actually used in the handler:

```
const handler: (options: Options) => Promise<void> = async ({
  apiToken,
  sessionId,
  appUrl,
  devTools,
  shiftTime,
  networkStubbing,
  moveBeforeClick,
  cookiesFile,
  disableRemoteFonts,
}) => {
```

```
export const COMMON_REPLAY_OPTIONS = {
  headless: OPTIONS.headless,
  devTools: OPTIONS.devTools,
  bypassCSP: OPTIONS.bypassCSP,
  padTime: OPTIONS.padTime,
  shiftTime: OPTIONS.shiftTime,
  networkStubbing: OPTIONS.networkStubbing,
  skipPauses: OPTIONS.skipPauses,
  disableRemoteFonts: OPTIONS.disableRemoteFonts,
};
```

It's probably a bit misleading to have them show in the CLI help but not actually be wired through.

Maybe for now do:

```
  devTools: OPTIONS.devTools,
  shiftTime: OPTIONS.shiftTime,
  networkStubbing: OPTIONS.networkStubbing,
  disableRemoteFonts: OPTIONS.disableRemoteFonts,
```

Can then always separately wire through more options.

Que3216 (approved):
One suggested change, but otherwise looks great!

PR #151, opened 2022-11-04
Pass original session start URL through to network event stubbing

Que3216:
Pass original session start URL through to network event stubbing so that it can swap out the origin correctly with the origin that matches the original recorded requests. Currently the origin swapping is broken, because it swaps out the `--appUrl` origin for the `--appUrl` origin.

### DIFF
--- a/packages/replayer/src/replay.utils.ts
+++ b/packages/replayer/src/replay.utils.ts
@@ -148,49 +148,60 @@ export const patchDate: (options: {
   `);
 };
 
-const getAppUrl: (options: {
-  sessionData: any;
+export const getStartUrl: (options: {
+  originalSessionStartUrl: URL;
   appUrl: string | null;
-}) => string = ({ sessionData, appUrl }) => {
+}) => string = ({ originalSessionStartUrl, appUrl }) => {
   if (!appUrl) {
-    const { startUrl, startURL } = sessionData.userEvents.window;
-    return startUrl || startURL;
+    return originalSessionStartUrl.toString();
   }
+
+  const parsedAppUrl = parseAppUrl(appUrl);
+
+  const appUrlIsJustOrigin =
+    parsedAppUrl.pathname === "/" && !parsedAppUrl.search && !parsedAppUrl.hash;
+
+  if (!appUrlIsJustOrigin) {
+    return parsedAppUrl.toString();
+  }
+
+  // Use the original URL's path, query string etc., but swap out the origin for the provided one
+  const mergedUrl = new URL(originalSessionStartUrl.toString());
+
+  mergedUrl.host = parsedAppUrl.host;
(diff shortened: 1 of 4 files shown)


packages/replayer/src/replay.utils.ts
```diff
@@ -148,49 +148,60 @@ export const patchDate: (options: {
   `);
 };
 
-const getAppUrl: (options: {
-  sessionData: any;
+export const getStartUrl: (options: {
```
Que3216, line 151: There is one change in the logic here.

There are two recorded start URLs -- sessionData.userEvents.window.startUrl is the URL after 1500ms, while session.startUrl is the URL straight away.

Previously if the appUrl was falsey and `session.startUrl` is truthy, then it'd use `sessionData.userEvents.window.startUrl` if `sessionData.userEvents.window.startUrl` was more than just an origin, and `session.startUrl` with the origin of `sessionData.userEvents.window.startUrl` otherwise. I'm pretty sure this was unintended / a bug.

Now if appUrl is falsey and `session.startUrl` is truthy then it'll use `session.startUrl`

packages/replayer/src/replayer.ts
```diff
@@ -121,14 +126,14 @@ export const replayEvents: ReplayEventsFn = async (options) => {
...
-        startUrl: string;
+        originalSessionStartUrl: string;
         onTimelineEvent: OnReplayTimelineEventFn;
       }) => Promise<void>;
     await setupReplayNetworkStubbing({
       page,
       logLevel,
       sessionData,
-      startUrl,
+      originalSessionStartUrl: originalSessionStartUrl.toString(),
```
Que3216, line 136: For back compat I'll make the meticulous side do `originalSessionStartUrl ?? startUrl // CLI versions prior to https://github.com/alwaysmeticulous/meticulous-sdk/pull/151/files send through a startUrl instead`.

alexivanov (approved):
:shipit:

packages/replayer/src/replayer.ts
Que3216: Actually realized I do still need the resolved startUrl. I'll pass both.

packages/replayer/src/replayer.ts
```diff
@@ -213,6 +223,10 @@ export const replayEvents: ReplayEventsFn = async (options) => {
     }
   }
 
+  logger.debug("Collecting tracing data...");
+  await page.tracing.stop();
+  logger.debug("Collected tracing data");
+
```
Que3216, line 229:
```suggestion
```

packages/replayer/src/replayer.ts
```diff
@@ -1,3 +1,4 @@
+import { join } from "path";
```
Que3216, line 1:
```suggestion
```

PR #153, opened 2022-11-04
Use replay v2 snippets in simulation debugger

alexivanov:
### Changes

This PR updates the `replay-debugger` to use replay v2 snippets, following similar logic to the standard replayer. The control plane in the debugger was already within the node process, so not much of a change in terms of user interaction. The browser context seeding and network stubbing logic is moved to the node process.

Fixes SIM-322


Que3216 (approved):

leafty (approved):
LGTM :tada: :clap:  It feels really nice to see the replay rewrite has made all simulation pieces cleaner and simpler

PR #157, opened 2022-11-07
Wire through moveBeforeClick and default it to true

Que3216:


packages/replayer/src/replayer.ts
```diff
@@ -191,7 +192,7 @@ export const replayEvents: ReplayEventsFn = async (options) => {
     page,
     logLevel,
     sessionData,
-    moveBeforeClick: true,
```
Que3216, line 194: I didn't realise, but this was actually always set to true in replay next (https://github.com/alwaysmeticulous/meticulous-sdk/commit/ffa0f0c1851800f94721427432638446fb66348e#diff-78f936c62eb1e7311e4a99f1c66a83ca0ce49f561d7d177189789aeac52ef06f).

This means it being set to false may not having been the thing causing the hover flakes (possible setting it to false may actually help if it has always been set to true, or we may need to move mouse to corner or something to prevent any hover states).

leafty: Yes, it is most likely my mistake :pray:

leafty (approved):

PR #160, opened 2022-11-09
Update replay options to support `baseSimulationId`

leafty:
This PR updates `ReplayEventsOptions` to accept `baseSimulationId` as a parameter.

This will be used to keep track of the base simulation to be used for generating a storyboard.


Que3216 (approved):
Code looks good. How are you planning on using baseSimulationId? Is the intent that replay-next-node would download the screenshots archive from the base simulation, and use that to compute which points to screenshot at? If so, could we have replay.command do this, and pass through the list of screenshot points to replayEvents?

leafty:
Marking as obsolete.

PR #161, opened 2022-11-09
Install virtual event loop before rest of page loads

Que3216:
If virtual time is enabled then we need to install the virtual event loop before any application code runs, since that code may snapshot copies of the setTimeout etc. functions

### DIFF
--- a/packages/replayer/src/replay.utils.ts
+++ b/packages/replayer/src/replay.utils.ts
@@ -4,7 +4,10 @@ import {
   ReplayEventsDependencies,
   SessionData,
 } from "@alwaysmeticulous/common";
-import { OnReplayTimelineEventFn } from "@alwaysmeticulous/sdk-bundles-api";
+import {
+  OnReplayTimelineEventFn,
+  VirtualTimeOptions,
+} from "@alwaysmeticulous/sdk-bundles-api";
 import log from "loglevel";
 import { DateTime, Duration } from "luxon";
 import { BrowserContext, Page, Viewport } from "puppeteer";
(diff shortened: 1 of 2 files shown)


packages/replayer/src/replay.utils.ts
```diff
@@ -105,6 +116,16 @@ export const createReplayPage: (options: {
   return page;
 };
 
+const installVirtualEventLoop = (page: Page) =>
+  page.evaluateOnNewDocument(`
+    const installVirtualEventLoop = window.__meticulous?.replayFunctions?.installVirtualEventLoop;
+    if (installVirtualEventLoop) {
+      installVirtualEventLoop();
+    } else {
+      console.error("Could not install virtual event loop since window.__meticulous.replayFunctions.installVirtualEventLoop was null or undefined");
```
leafty, line 125: Maybe this should throw an error?

packages/replayer/src/replay.utils.ts
```diff
@@ -92,6 +97,12 @@ export const createReplayPage: (options: {
   );
   await page.evaluateOnNewDocument(userInteractionsFile);
 
+  // If virtual time is enabled then we need to install the virtual event loop before
+  // any customer code runs, since that code may snapshot copies of the setTimeout etc. functions
```
leafty, line 101: I think I would avoid leaving comments like this in the open

packages/replayer/src/replay.utils.ts
Que3216: Throwing an error wouldn't do anything other than log to the developers console, because this is evaluated in global context on document load.

Do we currently capture and log uncaught exceptions during replay?

packages/replayer/src/replay.utils.ts
Que3216, line 101:
```suggestion
```

Que3216: Chatted on Slack to understand reasoning: reasoning was that comment adds bloat & not required.

leafty (approved):

PR #164, opened 2022-11-11
Update the simulate command to send diff data to Meticulous

leafty:
Update relevant CLI commands to send screenshot diff data to the Meticulous API.

### Details

* Diffing screenshots now computes the outcome for all screenshots, so that we can persist the diff data
* When `run-all-tests` is run, the `testRunId` is passed around to be tracked when the screenshot diff data is pushed to the Meticulous API

[SIM-334](https://toil.kitemaker.co/qpgVHi-Meticulous/pPGwrM-Engineering/items/334)

### DIFF
--- a/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
+++ b/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
@@ -163,12 +163,12 @@ export const checkScreenshotDiffResult = ({
 }): void => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
 
-  const missingHeadImagesResults = results.filter(
-    ({ outcome }) => outcome === "missing-head"
-  ) as ScreenshotDiffResultMissingHead[];
+  const missingHeadImagesResults = results.flatMap((result) =>
+    result.outcome === "missing-head" ? [result] : []
+  );
   if (missingHeadImagesResults.length) {
     const message = `Head replay is missing screenshots: ${missingHeadImagesResults
-      .map(({ baseScreenshotFile }) => baseScreenshotFile)
+      .map(({ baseScreenshotFile }) => basename(baseScreenshotFile))
       .sort()} => FAIL!`;
     logger.info(message);
     throw new ScreenshotDiffError(message, {
@@ -177,9 +177,9 @@ export const checkScreenshotDiffResult = ({
     });
   }
 
-  const missingBaseImagesResults = results.filter(
-    ({ outcome }) => outcome === "missing-base"
-  ) as ScreenshotDiffResultMissingBase[];
+  const missingBaseImagesResults = results.flatMap((result) =>
+    result.outcome === "missing-base" ? [result] : []
+  );
   if (missingHeadImagesResults.length) {
     const message = `Notice: Base replay is missing screenshots: ${missingBaseImagesResults
       .map(({ headScreenshotFile }) => basename(headScreenshotFile))
@@ -202,8 +202,8 @@ export const checkScreenshotDiffResult = ({
     }
 
     if (outcome === "diff" || outcome === "no-diff") {
-      const mismatch = Math.round(result.mismatchFraction * 100);
-      const threshold = Math.round(diffOptions.diffThreshold * 100);
+      const mismatch = (result.mismatchFraction * 100).toFixed(3);
+      const threshold = (diffOptions.diffThreshold * 100).toFixed(3);
       const message = `${mismatch}% pixel mismatch for screenshot ${basename(
         result.headScreenshotFile
       )} (threshold is ${threshold}%) => ${
--- a/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
+++ b/packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
@@ -188,16 +188,6 @@ export const checkScreenshotDiffResult = (
       throw new Error(message);
     }
 
-    // logComparisonResultMessage(
-    //   logger,
-    //   `${Math.round(
-    //     result.comparisonResult.mismatchFraction * 100
-    //   )}% pixel mismatch for screenshot ${basename(
-    //     result.headScreenshotFile
-    //   )} (threshold is ${Math.round(diffThreshold * 100)}%)`,
-    //   result.outcome
-    // );
-
     if (outcome === "diff" || outcome === "no-diff") {
       const mismatch = Math.round(result.mismatchFraction * 100);
       const threshold = Math.round(diffOptions.diffThreshold * 100);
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -18,199 +22,212 @@ import {
...
-        logger.info(
-          `Screenshot ${screenshotFileName} not present in base replay`
-        );
-        continue;
-      }
+  const headDiffResults = (
+    await Promise.all(
+      headReplayScreenshots.map(
+        async (screenshotFileName): Promise<ScreenshotDiffResult[]> => {
+          const identifier = getScreenshotIdentifier(screenshotFileName);
```
Que3216, line 69: [nit / not important] Could also pull this out as a function below if you want to reduce nesting.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -18,199 +22,212 @@ import {
...
+                width: baseScreenshot.width,
+                height: baseScreenshot.height,
+                mismatchPixels: comparisonResult.mismatchPixels,
+                mismatchFraction: comparisonResult.mismatchFraction,
+              },
+            ];
+          } catch (error) {
+            if (
+              error instanceof Error &&
+              error.message.startsWith("Cannot handle different size")
```
Que3216, line 128: [nit / not important] May be cleaner to compare the sizes ourselves to check this, or to have compareImages return a type union rather than throw an error, or use a shared constant or custom error type. That way this wouldn't break if we change the text of their error message slightly / non-obvious in compareImages that changing that text may break things.

leafty: Done.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
leafty: I would have to pass around 6 parameters or so, so I will leave it like this.

Que3216: Ack, that makes sense. The other trick I sometimes do in that kind of case if I want to reduce nesting is to pull it out as a function but within the other function:

```ts
  const diffWithCorrespondingBaseScreenshot = (headScreenshotFileName: string): Promise<ScreenshotDiffResult[]> => {

      // ... long function body

  };

  const headDiffResults = await Promise.all(headReplayScreenshots.map(diffWithCorrespondingBaseScreenshot).flat()));
```

That way it's the function has a name, so it's easier to understand what it's doing (you don't have to read the whole function body to get the gist of what it's doing in order to understand the rest of the code); the .flat() becomes more visible, rather than hidden at the bottom; and there's less tabbing, which often makes it a little easier to follow.

(other way of course is to use an options interface, but that can be a pain)

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -18,199 +24,244 @@ import {
...
+  baseReplayId: string;
+  headReplayId: string;
+  results: ScreenshotDiffResult[];
+  diffOptions: ScreenshotDiffOptions;
+}): void => {
+  const logger = log.getLogger(METICULOUS_LOGGER_NAME);
+
+  const missingHeadImagesResults = results.filter(
+    ({ outcome }) => outcome === "missing-head"
+  ) as ScreenshotDiffResultMissingHead[];
```
Que3216, line 174: [nit -- but worthwhile for future discriminated type union filters since it can occasionally reduce bugs introduced in refactors]

Note if you want to do this type safe, without a cast, you can do:

```ts
const missingHeadImagesResults = results.flatMap(result => result.outcome === "missing-head" ? [result] : []);
```

Typescript will then check the types for you (with current code it won't check that 'ScreenshotDiffResultMissingHead' aligns with the type implied by checking `result.outcome === "missing-head"`).

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -18,199 +24,244 @@ import {
...
-        )} (threshold is ${Math.round(diffThreshold * 100)}%)`,
-        result.outcome
-      );
-    });
+  const missingBaseImagesResults = results.filter(
+    ({ outcome }) => outcome === "missing-base"
+  ) as ScreenshotDiffResultMissingBase[];
+  if (missingHeadImagesResults.length) {
+    const message = `Notice: Base replay is missing screenshots: ${missingBaseImagesResults
+      .map(({ headScreenshotFile }) => basename(headScreenshotFile))
```
Que3216, line 191: Note: we use basename here but not in the missing-head case

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -18,199 +24,244 @@ import {
...
+      logger.info(message);
+      throw new ScreenshotDiffError(message, {
+        baseReplayId,
+        headReplayId,
+      });
+    }
+
+    if (outcome === "diff" || outcome === "no-diff") {
+      const mismatch = Math.round(result.mismatchFraction * 100);
+      const threshold = Math.round(diffOptions.diffThreshold * 100);
```
Que3216, line 212: [nit] I'd probably do `(result.mismatchFraction * 100).toFixed(3)` and `(diffOptions.diffThreshold * 100).toFixed(3)` since I think we'll sometimes use fraction thresholds that are < 0.01 (in current CI tests use diffThreshold of 0, and I've often seen diffs of about 0.01% of pixels, and it can be useful to know how many pixels differ -- currently [before this PR] it prints out something like "0% pixel mismatch, threshold is 0% -- FAIL" which isn't so useful).

Que3216 (approved):
Nice! 🚀

Very excited for this.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
Que3216: Same with below .filter of course.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
leafty: Done in c8ca30dd7105121e118b5a532208a8c8f75fc947

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
leafty: Done in c07f118

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
leafty: Done in c07f118

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
leafty: Done in c07f118

PR #166, opened 2022-11-11
Pass through session start time to virtual event loop

Que3216:
This will ensure the values returned by `Date.now()` etc. are always _exactly_ the same on every skip pauses run -- while currently there's non-determinism due the time elapsed from evaluating `patchDate` and evaluating `installVirtualEventLoop` (the later which calls Date.now() on the patched date to work out the start time).

Once this has merged I'll update installVirtualEventLoop to use the passed in timestamp, and to extend from `window.__meticulous.timeshift.OriginalDate` if it exists _and_ if the new sessionStartTime param is being passed through, rather than from `window.Date` (which would shift twice). Since I can do this as a single change there won't be any issues with API breaks (old CLI, new bundle etc.).

### DIFF
--- a/packages/replayer/src/replay.utils.ts
+++ b/packages/replayer/src/replay.utils.ts
@@ -7,6 +7,7 @@ import {
 import {
   OnReplayTimelineEventFn,
   VirtualTimeOptions,
+  InstallVirtualEventLoopOpts,
 } from "@alwaysmeticulous/sdk-bundles-api";
 import log from "loglevel";
 import { DateTime, Duration } from "luxon";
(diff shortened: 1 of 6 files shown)


packages/replayer/src/replay.utils.ts
```diff
@@ -114,15 +116,31 @@ export const createReplayPage: (options: {
...
+      const installVirtualEventLoop = window.__meticulous?.replayFunctions?.installVirtualEventLoop;
+      if (installVirtualEventLoop) {
+        installVirtualEventLoop(${JSON.stringify(opts)});
+      } else {
+        console.error("Could not install virtual event loop since window.__meticulous.replayFunctions.installVirtualEventLoop was null or undefined");
+      }
+  `);
+};
+
+export const getSessionStartTime = (sessionData: SessionData): DateTime =>
```
Que3216, line 133: Because this is exported from the package want to give it a name that doesn't assume it's calculated from rrweb data, but instead just uses whatever is our current canonical way of calculating the session start time, as expected by patchDate's and installVirtualEventLoop's sessionStartTime param

leafty (approved):
LGTM

PR #168, opened 2022-11-14
Add stronger shared types for SessionData and ReplayTimelineEntry

Que3216:
I defined these types by checking a combination of the code and the actual data.

PR #169, opened 2022-11-14
Update date shifting to not rely on rrweb

Que3216:
We automatically abandon recording rrweb data if it's too large since it's generally not essential for replay. However there is one remaining place where it is essential for replay and that's for when we do date shifting. This PR removes that dependency by falling back to using the user events to calculate the session start and session duration.

This will fix the issue for now, and in the meantime we can work on adding a dedicated session start timestamp (that event.timeStamps are relative to) that's measured in Date.now() time rather than performance.timeOrigin time, to ensure we can reproduce the original Date.now() calls exactly.

### DIFF
--- a/packages/replayer/src/replay.utils.ts
+++ b/packages/replayer/src/replay.utils.ts
@@ -136,18 +136,28 @@ export const getSessionStartTime = (sessionData: SessionData): DateTime => {
     return rrWebTimeRange.min;
   }
 
-  // Note that the user event timestamp is based on performance.timeOrigin & the offset on the raw event,
-  // which can differ from the time returned by Date.now(). However for old sessions we don't record Date.now()
-  // time
-  const minUserEventTimestamp =
-    sessionData.userEvents?.event_log[0]?.timeStampRaw;
-  if (minUserEventTimestamp == null) {
+  const userEventEventLog = sessionData.userEvents?.event_log ?? [];
+  if (userEventEventLog.length === 0) {
     const logger = log.getLogger(METICULOUS_LOGGER_NAME);
     logger.warn(
       "No user or rrweb events, so cannot accurately determine start timestamp. Using current Date instead, however this will be unstable"
     );
     return DateTime.now().toUTC();
   }
+
+  // event.timeStamp differs from event.timeStampRaw in that (a) it is relative to the start of the session,
+  // (b) it has been adjusted to subtract the time between constructing the recorder and calling.start().
+  //
+  // By subtracting event.timeStamp from event.timeStampRaw we get the performance.timeOrigin plus the
+  // the time between constructing the recorder and calling.start(), in other words: the time at which recording
+  // started.
+  //
+  // Note that performance.timeOrigin uses monotonic clock time, which can differ by multiple hours
+  // from the time returned by Date.now(). We may in future want to record a dedicated
+  // session start time to ensure we fully accuractly recreate the original values returned by the
+  // Date.now() calls.
+  const minUserEventTimestamp =
+    userEventEventLog[0]?.timeStampRaw - userEventEventLog[0]?.timeStamp;
   return DateTime.fromMillis(minUserEventTimestamp).toUTC();
 };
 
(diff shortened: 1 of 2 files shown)


packages/replayer/src/replay.utils.ts
```diff
@@ -130,31 +130,84 @@ const installVirtualEventLoop = (page: Page, sessionStartTime: DateTime) => {
...
+    );
+    return DateTime.now().toUTC();
+  }
+
+  // event.timeStamp differs from event.timeStampRaw in that (a) it is relative to the start of the session,
+  // (b) it has been adjusted to subtract the time between constructing the recorder and calling.start().
+  //
+  // By subtracting event.timeStamp from event.timeStampRaw we get the performance.timeOrigin plus the
+  // the time between constructing the recorder and calling.start(), in other words: the time at which recording
+  // started.
```
leafty, line 153: I would move this explainer to jsdoc on `timeStamp` and `timeStampRaw` fields.

packages/replayer/src/replay.utils.ts
```diff
@@ -130,31 +130,84 @@ const installVirtualEventLoop = (page: Page, sessionStartTime: DateTime) => {
...
   sessionData: SessionData
-) => [DateTime, DateTime] = (sessionData) => {
+): { min: DateTime; max: DateTime } | null => {
   const rrwebTimestamps = (sessionData.rrwebEvents as { timestamp?: number }[])
     .map((event) => event.timestamp || NaN)
     .filter((ts) => !isNaN(ts));
+
+  if (rrwebTimestamps.length === 0) {
+    return null;
+  }
```
leafty, line 202: `luxon` correctly propagates invalid timestamp computations, so this is not strictly needed

packages/replayer/src/replay.utils.ts
Que3216: Yep that's a good call. Most of the bottom half will remaining, but the top half is much better suited to be on the fields. That will depend on https://github.com/alwaysmeticulous/meticulous-sdk/pull/168 merging I think.

Que3216: Let's get this one merged, and then can clean up as part of or after 168.

packages/replayer/src/replay.utils.ts
Que3216: It's correctly that the code could be made to work without this. You'd get DateTime.fromMillis of Infinity which would return invalid timestamps, which you could then check in getSessionDuration. And that's how I initially implemented it, however I refactored to this to ensure it's explicit -- the reader can understand that we intentionally handle the empty case in a particular way, without having to (a) know the behaviour of Math.max on an empty set, (b) the behaviour of luxon on infinite values, (c) that we wrote the code intentionally taking advantage of these behaviours, rather than accidentally, to handle a special case of the rrwebTimestamps array being empty.

Que3216: General rule here is that code that is quicker and easier to understand the programmers intent and the intended flows of the code / different cases is better even if the code has more lines/characters.

packages/replayer/src/replay.utils.ts
leafty: :+1: alright

packages/replayer/src/replay.utils.ts
```diff
@@ -130,31 +130,84 @@ const installVirtualEventLoop = (page: Page, sessionStartTime: DateTime) => {
   `);
 };
 
-export const getSessionStartTime = (sessionData: SessionData): DateTime =>
-  getMinMaxRrwebTimestamps(sessionData)[0];
+export const getSessionStartTime = (sessionData: SessionData): DateTime => {
+  const rrWebTimeRange = getMinMaxRrwebTimestamps(sessionData);
+  if (rrWebTimeRange != null) {
```
leafty, line 135:
```suggestion
  if (rrWebTimeRange != null && rrWebTimeRange.min.isValid) {
```

packages/replayer/src/replay.utils.ts
leafty: :+1:

leafty (approved):

packages/replayer/src/replay.utils.ts
Que3216: Ah yeah good call

PR #170, opened 2022-11-16
Refactor replay options

leafty:


packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -296,8 +294,9 @@ export const replayCommandHandler = async ({
       });
     } catch (error) {
       if (exitOnMismatch) {
-        throw error;
+        process.exit(1);
       }
+      throw error;
```
leafty, line 299: Note: this was a bug; it was swallowing screenshot diff failures.

Que3216 (approved):
Nice!

PR #171, opened 2022-11-16
Add error case to replay results

Que3216:
This should make debugging easier, because it means that if an error is thrown it still writes and updates the output. Presumably the end state screenshot will still differ, so we'll still get a failed comparison. If it's a passive session, and the storyboard screenshots are all there and all identical then it may not be important to the user that the replay failed after the last storyboard screenshot (depending on why it failed), though I expect at some point we will want to surface this.

### DIFF
--- a/packages/replayer/src/replayer.ts
+++ b/packages/replayer/src/replayer.ts
@@ -196,39 +196,49 @@ export const replayEvents: ReplayEventsFn = async (options) => {
       ? { enabled: true, screenshotsDir }
       : { enabled: false };
   const sessionDuration = getSessionDuration(sessionData);
-  const replayResult = await replayUserInteractions({
-    page,
-    logLevel,
-    sessionData,
-    moveBeforeClick,
-    virtualTime,
-    storyboard,
-    onTimelineEvent,
-    ...(sessionDuration != null
-      ? { sessionDurationMs: sessionDuration?.milliseconds }
-      : {}),
-    ...(maxDurationMs != null ? { maxDurationMs } : {}),
-    ...(maxEventCount != null ? { maxEventCount } : {}),
-  });
-  logger.debug(`Replay result: ${JSON.stringify(replayResult)}`);
+  try {
+    const replayResult = await replayUserInteractions({
+      page,
+      logLevel,
+      sessionData,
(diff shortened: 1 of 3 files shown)


packages/replayer/src/replayer.ts
```diff
@@ -196,39 +196,49 @@ export const replayEvents: ReplayEventsFn = async (options) => {
...
+      logger.debug(`Padtime: ${timeToPad} ${sessionDuration.toISOTime()}`);
+      if (timeToPad > 0) {
+        await new Promise<void>((resolve) => {
+          setTimeout(() => {
+            resolve();
+          }, timeToPad);
+        });
+      }
     }
+  } catch (error: unknown) {
```
leafty, line 234: Do we need the explicit type here? I would always assume this to be `unknown`.

packages/replayer/src/replayer.ts
```diff
@@ -271,3 +281,15 @@ const shouldHoldBrowserOpen = () => {
...
+  error: unknown
+): { message: string | null; stack: string | null } => {
+  if (error == null) {
+    return { message: null, stack: null };
+  }
+  return {
+    message: (error as any).message ?? (error as any).toString?.() ?? null,
+    stack: (error as any).stack ?? null,
+  };
+};
```
leafty, line 295: This one could be unsafe if the object thrown is not an `Error`, e.g. (`{ message: number }`).

```suggestion
const serializeError = (
  error: unknown
): { message: string | null; stack: string | null } => {
  if (error == null) {
    return { message: null, stack: null };
  }
  const message =
    toStringOrNull((error as any).message) ?? toStringOrNull(error);
  const stack = toStringOrNull((error as any).stack);
  return {
    message,
    stack,
  };
};

const toStringOrNull = (value: unknown): string | null =>
  value != null ? `${value}` : null;
```

leafty (approved):
LGTM :clap:

packages/replayer/src/replayer.ts
Que3216, line 234:
```suggestion
  } catch (error) {
```

Yep it does default to unknown (on old versions of typescript it'd default to any).

PR #172, opened 2022-11-17
Load assets once when running N tests in parallel, rather than N times

Que3216:
We're intermittently getting failures like:

```
[run-all-tests] Starting simulation...
[run-all-tests] 
[run-all-tests] ERROR: TypeError: replayUserInteractions is not a function
[run-all-tests] ERROR:     at replayEvents (/home/runner/work/meticulous/meticulous/node_modules/@alwaysmeticulous/replayer/dist/replayer.js:107:11)
[run-all-tests] ERROR:     at async replayCommandHandler (/home/runner/work/meticulous/meticulous/node_modules/@alwaysmeticulous/cli/dist/commands/replay/replay.command.js:117:5)
[run-all-tests] ERROR:     at async handleReplay (/home/runner/work/meticulous/meticulous/node_modules/@alwaysmeticulous/cli/dist/deflake-tests/deflake-tests.handler.js:14:20)
[run-all-tests] ERROR:     at async deflakeReplayCommandHandler (/home/runner/work/meticulous/meticulous/node_modules/@alwaysmeticulous/cli/dist/deflake-tests/deflake-tests.handler.js:35:25)
[run-all-tests] ERROR:     at async main (/home/runner/work/meticulous/meticulous/node_modules/@alwaysmeticulous/cli/dist/parallel-tests/task.handler.js:46:20)
[run-all-tests] ERROR: 
[run-all-tests] Simulation done!
[run-all-tests] 
```

And:

```
[run-all-tests] ERROR: TypeError: setupReplayNetworkStubbing is not a function
```

(more details in SIM-308)

I think it's likely that this is because of a race condition in loading the assets. If you start 10 tests at once in 10 threads, then they'll all load assets at the same time (etag in metadata only updated after asset finishes loading), and try writing to the same files. This PR pulls the asset loading up, so it only happens once before running any of the tests, instead of one per test. If this doesn't solve the issue (we still see such failures) then at least it will have made the code more performant.

Testing performed:
 - run-all-tests still works
 - replay still works


packages/cli/src/local-data/replay-assets.ts
```diff
@@ -84,3 +101,36 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   await saveAssetMetadata(assetMetadata);
   return filePath;
 };
+
+const getOrCreateAssetsDir: () => Promise<string> = async () => {
```
Que3216, line 105: No change to these functions. They've just been moved to the bottom, as per step-down rule, and the 'export' prefix removed (they're only used locally in this file).

packages/cli/src/local-data/replay-assets.ts
```diff
@@ -3,56 +3,73 @@ import { basename, join } from "path";
...
+      "replay/v2/node-browser-context.bundle.js"
+    );
+    const nodeNetworkStubbing = await fetchAsset(
+      "replay/v2/node-network-stubbing.bundle.js"
+    );
+    const nodeUserInteractions = await fetchAsset(
+      "replay/v2/node-user-interactions.bundle.js"
+    );
 
-export const loadAssetMetadata: () => Promise<AssetMetadata> = async () => {
```
Que3216, line 30: No change to these functions, they've just been moved down to bottom and the export removed.

packages/replayer/src/replayer.ts
```diff
@@ -129,17 +128,10 @@ export const replayEvents: ReplayEventsFn = async (options) => {
   // Set-up network stubbing if required
   if (networkStubbing) {
     // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const networkStubbingModule = require(dependencies.nodeNetworkStubbing
+    const networkStubbingExports = require(dependencies.nodeNetworkStubbing
```
leafty, line 131: Nitpick: why rename it? I see it as an ES or CommonJS module since it is the result of calling `require`.

leafty (approved):
LGTM

Good idea!

packages/replayer/src/replayer.ts
Que3216: It wasn't really intentional rename / I don't have any real preference between the two names. I refactored it out into a function and then back again when I changed my mind, and so ended up with a different name.

PR #173, opened 2022-11-18
Move types that are persisted in the database to @alwaysmeticulous/api

Que3216:
Additional constraints apply to changing types that are stored in the db/s3, so want to separate such types out. This also means we can cut any dependencies from the non-bundle packages in the main Meticulous repo on sdk-bundles-api.


packages/api/README.md
```diff
@@ -1,3 +1,27 @@
 # Meticulous API
 
-This packages contains type definitions used by the Meticulous API.
+This packages contains type definitions that are persisted in the database or S3 and also used in the SDK.
```
leafty, line 3: :thinking: I think reading this README as an external user may be confusing.

leafty (approved):

packages/api/README.md
Que3216: Yeah, I think that's OK though. We'll want to carefully handhold any changes by external users to this package, since changes here change what's stored in our DBs.

PR #174, opened 2022-11-18
Add baseCommitSha argument

Que3216:
This should enable a workflow where the baseReplayIds are not checked into your meticulous.json file. So if the diffs are intended you can just merge the PR, without having to update your meticulous.json file. The intent is that this will be used as the recommended method for both active and passive tests. The --baseCommitSha will be automatically passed in by a github action that wraps the Meticulous CLI.

### DIFF
--- a/packages/cli/src/utils/run-all-tests.utils.ts
+++ b/packages/cli/src/utils/run-all-tests.utils.ts
@@ -61,11 +61,7 @@ export const getTestsToRun = async ({
     (testCase) => testCase.baseReplayId == null
   );
   const testCasesWithBaseReplayId = uncachedTestCases.flatMap(
-    (testCase): TestCase[] =>
-      // Note: explictly setting the baseReplayId is required for typescript to be happy with types
-      testCase.baseReplayId == null
-        ? []
-        : [{ ...testCase, baseReplayId: testCase.baseReplayId }]
+    (testCase): TestCase[] => (testCase.baseReplayId == null ? [] : [testCase])
   );
 
   if (testCasesMissingBaseReplayId.length === 0) {
@@ -89,8 +85,7 @@ export const getTestsToRun = async ({
 
   return uncachedTestCases.flatMap((test) => {
     if (test.baseReplayId != null) {
-      // Note: explictly setting the baseReplayId is required for typescript to be happy with types
-      return [{ ...test, baseReplayId: test.baseReplayId }];
+      return [test];
     }
     const baseReplayId = baseReplayIdBySessionId[test.sessionId];
     if (baseReplayId == null) {
@@ -98,7 +93,7 @@ export const getTestsToRun = async ({
       logger.warn(
         `Skipping comparisons for test "${test.title}" since no result to compare against stored for base commit ${baseCommitSha}`
       );
-      return test;
+      return [test];
     }
     return [{ ...test, baseReplayId }];
   });
--- a/packages/cli/src/utils/run-all-tests.utils.ts
+++ b/packages/cli/src/utils/run-all-tests.utils.ts
@@ -115,16 +115,6 @@ const getBaseReplayIdsBySessionId = async ({
     client,
     commitSha: baseCommitSha,
   });
-
-  if (baseTestRun?.status === "Running") {
-    // Note: for now we just ignore results that haven't completed yet, but in future we may want to block
-    // on the test run completely (poll until it's complete, or timed out).
-    const logger = log.getLogger(METICULOUS_LOGGER_NAME);
-    logger.warn(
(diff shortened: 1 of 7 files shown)


packages/cli/src/utils/run-all-tests.utils.ts
```diff
@@ -39,3 +56,69 @@ export const getTestsToRun: (options: {
...
+};
+
+const getBaseReplayIdsBySessionId = async ({
+  client,
+  baseCommitSha,
+}: {
+  client: AxiosInstance;
+  baseCommitSha: string;
+}): Promise<Record<string, string>> => {
+  const baseReplays = await getCachedTestRunResults({
```
Que3216, line 114: This won't work for any users that have multiple run-all-tests commands setup on their CI, or are occasionally running run-all-tests locally. So I'll do a follow up PR after this to update the `/cache` endpoint to take some additional parameters. That could be a new param you pass via the CLI that uniquely identifies the test run comparison set, and the user provides, so that we can compare like for like (`--comparisonSet "main-ci-screenshots"`). Or maybe we pass to the /cache endpoint some information that helps it narrow down which test run to return, when there are multiple test runs for a commit, and it returns the closest matching one. Some possible info we could match on, each with their own tradeoffs:

 - Whether it's running on CI and the workflow name (https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)
 - The meticulous.json file path within the repo
 - The session ids, and the various parameters for the replay

Maybe the best solution would be:

 - We store whether a test run was run on CI or not.
 - The cache endpoint returns the most recent test run on CI if there are multiple test runs for the commit.
 - If users want to run multiple different sets of tests on CI simultaneously then they must provide a unique `--comparisonSet` or `--testSetId` parameter for each separate call to run-all-tests. Potentially this testSetId would default to the meticulous.json path within the repo.

leafty: Why not make an endpoint which allows to resolve: `(sessionId, commit) => latest simulation id`?

Que3216: Because there may be multiple replays for a session on the same commit run in different environments (localhost vs server; different settings - e.g. on our CI we run with/without --skipPauses etc.; eventually customers may want to run tests with dark mode on vs dark mode off etc.)

Que3216: Yep proposal would be (1) start out with this PR as is, so it risks matching wrong screenshot set if you also run run-all-tests locally on commits on your main branch (b) then update to doing the CI vs non-CI thing, and then (c) at some time later when we need it extend `test-runs/cache`/`test-runs/latest-run` to also support passing in the testSuiteId parameter.

packages/cli/src/utils/run-all-tests.utils.ts
```diff
@@ -39,3 +56,70 @@ export const getTestsToRun: (options: {
...
+
+  const baseReplayIdBySessionId = await getBaseReplayIdsBySessionId({
+    client,
+    baseCommitSha,
+  });
+
+  return uncachedTestCases.flatMap((test) => {
+    if (test.baseReplayId != null) {
+      // Note: explictly setting the baseReplayId is required for typescript to be happy with types
+      return [{ ...test, baseReplayId: test.baseReplayId }];
```
leafty, line 93:
```suggestion
      return [{ ...test }];
```

packages/cli/src/utils/run-all-tests.utils.ts
```diff
@@ -39,3 +56,70 @@ export const getTestsToRun: (options: {
...
+
+  const testCasesMissingBaseReplayId = uncachedTestCases.filter(
+    (testCase) => testCase.baseReplayId == null
+  );
+  const testCasesWithBaseReplayId = uncachedTestCases.flatMap(
+    (testCase): TestCase[] =>
+      // Note: explictly setting the baseReplayId is required for typescript to be happy with types
+      testCase.baseReplayId == null
+        ? []
+        : [{ ...testCase, baseReplayId: testCase.baseReplayId }]
```
leafty, line 68:
```suggestion
      testCase.baseReplayId == null ? [] : [{ ...testCase }]
```

leafty (approved):
LGTM

I'm not seeing the TypeScript errors.

packages/cli/src/utils/run-all-tests.utils.ts
Que3216: Ah good shout! This was actually out of date (I used to have a version of TestCase with the baseReplayId required and one without, and this was needed for converting from the later to the former, but it's no longer needed)

PR #175, opened 2022-11-21
Save bundle files with the .cjs extension

leafty:
This makes sure the bundle files are interpreted as CommonJS and not ESM.

See error:
```
Error [ERR_REQUIRE_ESM***: require() of ES Module /app/assets/node-network-stubbing.bundle.js from /app/node_modules/@alwaysmeticulous/replayer/dist/replayer.js not supported.
node-network-stubbing.bundle.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename node-network-stubbing.bundle.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /app/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

### DIFF
--- a/packages/cli/src/local-data/replay-assets.ts
+++ b/packages/cli/src/local-data/replay-assets.ts
@@ -82,7 +82,10 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const entry = assetMetadata.assets.find(
     (item) => item.fileName === assetFileName
   );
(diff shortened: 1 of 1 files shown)


Que3216 (approved):

leafty:
Note: this may break for client with existing assets, but only until new bundle versions are published.

PR #176, opened 2022-11-21
Add support for the --no-sandbox Chromium flag

leafty:


packages/cli/src/command-utils/common-options.ts
```diff
@@ -76,6 +76,11 @@ export const OPTIONS = {
     description: "Pass the disable remote fonts flag into chromium",
     default: false,
   },
+  noSandbox: {
+    boolean: true,
+    description: "Pass the no sandbox flag into chromium",
+    default: false,
+  },
```
Que3216, line 83: If we start passing a lot of these we may want some more generic `--chromiumFlags` option.

Que3216: But I think good like this for now. It does reduce the API surface / otherwise we wouldn't know what flags people may be overriding.

leafty: Yes, I also thought about doing that if we need to control many flags.

leafty: The question being how to properly handle the common options and potential overrides.

PR #178, opened 2022-11-23
Extract run all tests

Que3216:


packages/cli/src/commands/bootstrap/bootstrap.command.ts
```diff
@@ -26,7 +26,7 @@ const handler: (options: Options) => Promise<void> = async () => {
   });
```
Que3216, line 26: This refactor is in a seperate commit:

image

So may be easier to review that commit, and then review the other commits

packages/cli/src/parallel-tests/run-all-tests.ts
```diff
@@ -0,0 +1,248 @@
...
+  logger.info("Results");
+  logger.info("=======");
+  logger.info(`URL: ${testRunUrl}`);
+  logger.info("=======");
+  results.forEach(({ title, result }) => {
+    logger.info(`${title} => ${result}`);
+  });
+
+  if (githubSummary) {
+    await writeGitHubSummary({ testRunUrl, results });
```
Que3216, line 226: I'm planning to pull this out to live in the action in a later PR, and deprecate the githubSummary param.

packages/cli/src/parallel-tests/run-all-tests.ts
```diff
@@ -0,0 +1,248 @@
...
+        testRunId: testRun.id,
+        replayEventsDependencies,
+      });
+      results.push(result);
+      onTestFinished?.({
+        id: testRun.id,
+        url: testRunUrl,
+        status: "Running",
+        progress: {
+          ...getResultCounts(results),
```
Que3216, line 191: Note: this is N^2, but I think OK

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -152,6 +155,15 @@ export const runAllTestsInParallel: (
...
         results.push(result);
+        onTestFinished?.({
+          failedTestCases: results.filter(
+            (testCase) => testCase.result === "fail"
+          ).length,
+          passedTestCases: results.filter(
+            (testCase) => testCase.result === "pass"
+          ).length,
+          runningTestCases: queue.length + inProgress,
+        });
```
Que3216, line 166: Note: this is N^2, but I think OK. I may update later.

packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
```diff
@@ -67,7 +49,7 @@ const handler: (options: Options) => Promise<void> = async ({
   networkStubbing,
   githubSummary,
   parallelize,
-  parallelTasks,
+  parallelTasks: _parallelTasks,
```
leafty, line 52: Nitpick, convention:

```suggestion
  parallelTasks: parallelTasks_,
```

packages/cli/src/parallel-tests/run-all-tests.types.ts
```diff
@@ -0,0 +1,5 @@
+export interface TestRunProgress {
+  failedTestCases: number;
+  passedTestCases: number;
+  runningTestCases: number;
+}
```
leafty, line 5: Non-blocking: I am thinking this could also be similar data to what we push through with `putTestRunResults`.

leafty (approved):
LGTM

PR #179, opened 2022-11-24
Return screenshot diff results from run-all-tests

Que3216:
This is required for https://github.com/alwaysmeticulous/report-diffs-action/pull/26, which is important to ensure we don't report that a session failed with a link when the link leads the user to an empty set of screenshot diffs (e.g. because all sessions completely failed to even run), and since the # differing screens is more useful feedback than the # of differing sessions (the sessions are pretty random, some cover many distinct parts of the app) and lets the user know what to expect when they click the link.

### DIFF
--- a/packages/cli/src/api/test-run.api.ts
+++ b/packages/cli/src/api/test-run.api.ts
@@ -1,7 +1,8 @@
 import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
 import axios, { AxiosError, AxiosInstance } from "axios";
+import { restoreDefaultPrompts } from "inquirer";
 import log from "loglevel";
-import { TestCaseResult } from "../config/config.types";
+import { DetailedTestCaseResult, TestCaseResult } from "../config/config.types";
 
 export type TestRunStatus = "Running" | "Success" | "Failure";
 
@@ -78,7 +79,7 @@ export interface GetLatestTestRunResultsOptions {
 export const getCachedTestRunResults = async ({
   client,
   commitSha,
-}: GetLatestTestRunResultsOptions): Promise<TestCaseResult[]> => {
+}: GetLatestTestRunResultsOptions): Promise<DetailedTestCaseResult[]> => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
 
   if (!commitSha || commitSha === "unknown") {
@@ -91,7 +92,12 @@ export const getCachedTestRunResults = async ({
       ?.results ?? [];
 
   // Only return passing tests
-  return results.filter(({ result }) => result === "pass");
+  return (
+    results
+      .filter(({ result }) => result === "pass")
+      // If it passed it can't have any screenshot diffs:
+      .map((result) => ({ ...result, screenshotDiffResults: [] }))
+  );
 };
 
 export const getLatestTestRunResults = async ({
--- a/packages/cli/src/api/test-run.api.ts
+++ b/packages/cli/src/api/test-run.api.ts
@@ -78,7 +78,7 @@ export interface GetLatestTestRunResultsOptions {
 export const getCachedTestRunResults = async ({
   client,
   commitSha,
-}: GetLatestTestRunResultsOptions): Promise<DetailedTestCaseResult[]> => {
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -1,5 +1,9 @@
 import { mkdir, mkdtemp, writeFile } from "fs/promises";
```
Que3216, line 1: The diff in this file is a bit tricky to read because a bunch of code has been moved around. May be easier to compare before vs after.

packages/cli/src/commands/replay/utils/compute-and-save-diff.ts
```diff
@@ -0,0 +1,83 @@
...
+  baseReplayId,
+  tempDir,
+  headReplayId,
+  screenshottingOptions,
+  testRunId,
+}: ComputeAndSaveDiffOptions): Promise<{
+  screenshotDiffResults: ScreenshotDiffResult[];
+  screenshotDiffError: ScreenshotDiffError | null;
+}> => {
+  const logger = log.getLogger(METICULOUS_LOGGER_NAME);
```
Que3216, line 41: This is mostly just cut and pasted out from replay.command.ts (replay.command.ts was getting pretty big)

packages/cli/src/api/test-run.api.ts
```diff
@@ -91,7 +91,12 @@ export const getCachedTestRunResults = async ({
       ?.results ?? [];
 
   // Only return passing tests
-  return results.filter(({ result }) => result === "pass");
+  return (
+    results
+      .filter(({ result }) => result === "pass")
+      // If it passed it can't have any screenshot diffs:
+      .map((result) => ({ ...result, screenshotDiffResults: [] }))
```
leafty, line 98: Note: I think `ScreenshotDiffResultMissingBase` would still pass.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -158,7 +158,7 @@ export const checkScreenshotDiffResult = ({
   headReplayId: string;
   results: ScreenshotDiffResult[];
   diffOptions: ScreenshotDiffOptions;
-}): void => {
+}): ScreenshotDiffError | null => {
```
leafty, line 161: nitpick: to be even more explicit, the return type could be something like `ScreenshotMatch | ScreenshotDiffError`, or `"match" | ScreenshotDiffError`.

packages/cli/src/parallel-tests/run-all-tests.ts
```diff
@@ -215,7 +215,12 @@ export const runAllTests = async ({
     client,
     testRunId: testRun.id,
     status: overallStatus,
-    resultData: { results },
+    resultData: {
+      results: results.map(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        ({ screenshotDiffResults, ...coreResult }) => coreResult
+      ),
+    },
```
leafty, line 223: Discussion for later: we may want to update which data we store where (`Replay, TestRun, ReplayDiff`).

leafty (approved):
LGTM

packages/cli/src/api/test-run.api.ts
Que3216: Ah that's true :/. Hmm I could do a BE call to load the information.

packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
Que3216: Yep I like that. I'll do something like ScreenshotsMatch | ScreenshotsDoNotMatch, and change ScreenshotDiffError to be a ScreenshotsDoNotMatch `{ type: "error", message: ... }` kind of thing instead of a literal Error (we no longer actually throw it anywhere).

packages/cli/src/parallel-tests/run-all-tests.ts
Que3216: Yes -- or if the BE already has the full ReplayDiff etc., we could have the BE return this, and run-all-tests can pipe through in same format.

PR #181, opened 2022-11-25
Performance instrumentation

alexivanov:
### Changes

This adds performance instrumentation to the `simulate`(replay) command handler.


leafty (changes requested):
I think we should probably have `METICULOUS_TELEMETRY_DISABLED=1` to deactivate it.

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -13,6 +13,7 @@ import {
   ReplayTarget,
   StoryboardOptions,
 } from "@alwaysmeticulous/common";
+import * as Sentry from "@sentry/node";
```
Que3216, line 16: [nit, not important, feel free to ignore] I think it can sometimes help tree shaking, and also make it clearer to human readers which parts of Sentry we actually depend upon if you do:

```suggestion
import { startTransaction } from "@sentry/node";
```

packages/replayer/src/timing.ts
```diff
@@ -0,0 +1,33 @@
+import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
```
Que3216, line 1: This can be deleted now / is this still used?

packages/replayer/src/replayer.ts
```diff
@@ -264,6 +318,9 @@ export const replayEvents: ReplayEventsFn = async (options) => {
 
   logger.debug("Closing browser");
   await browser.close();
+
+  parentPerformanceSpan?.finish();
+  logger.info("Updated");
```
Que3216, line 323: Is this meant to stay, or just temp?

Que3216 (approved):

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -78,13 +79,19 @@ export const replayCommandHandler = async ({
   testRunId,
   replayEventsDependencies,
 }: ReplayOptions): Promise<ReplayResult> => {
+  const transaction = Sentry.startTransaction({
+    name: "replay.command_handler",
```
leafty, line 83: For nicer Sentry logs:

```suggestion
    name: "replay.command_handler",
    description: "Handle the replay command",
    op: "replay.command_handler",
```

leafty (approved):

PR #186, opened 2022-11-29
Retry on failure

Que3216:
Adds the new argument:

   **name**: maxRetiresOnFailure
   **description**: If set to a value greater than 0 then will re-run any replays that give a screenshot diff and mark them as a flake if the screenshot generated on one of the retryed replays differs from that in the first replay.

And stores such flakes in the backend as:

```ts
export interface ScreenshotDiffResultFlake {
  outcome: "flake";

  /**
   * The original diff. Can be any outcome but for 'no-diff'.
   */
  diffToBaseScreenshot: Omit<ScreenshotDiffResult, "identifier">;

  /**
   * The diffs created by retrying taking the head screenshot and comparing
   * it to the original head screenshot. At least one of these will have an
   * outcome other than no-diff, hence why this failure is marked as a flake.
   *
   * Note that in the context of these diffs base means the original head screenshot taken,
   * and head means the new head screenshot taken.
   */
  diffsToHeadScreenshotOnRetries: Array<
    Omit<
      ScreenshotDiffResult | { outcome: "missing-base-and-head" },
      "identifier"
    >
  >;
}
```

Currently the UI completely ignores flakes ([example test run](https://app.meticulous.ai/projects/meticulous/meticulous_internal_tests/test-runs/ttDhDDFCqHdm8GftMkgDpqDfth)), though after this PR I'll update it to provide a clearer message. Something like:

> **All 1 screens tested had zero reproducible differences**
> Meticulous replayed 1 user sessions before and after your PR and compared the 1 screenshots taken in each run. Differences were found for 1 of the screenshots, but when we retried taking those screenshots we got different results - so we're assuming the original difference was just a flake.

Or in the case there are some identical screenshots too, then:

> **All 1 screens tested had zero reproducible differences**
> Meticulous replayed 1 user sessions before and after your PR and compared the 2 screenshots taken in each run. Differences were found for 1 of the screenshots, but when we retried taking those screenshots we got different results - so we're assuming the original difference was just a flake.

### DIFF
--- a/packages/cli/src/parallel-tests/merge-test-results.ts
+++ b/packages/cli/src/parallel-tests/merge-test-results.ts
@@ -1,0 +1,97 @@
+import {
+  ScreenshotDiffResult,
+  ScreenshotIdentifier,
+} from "@alwaysmeticulous/api";
+import { logger } from "@sentry/utils";
+import { DetailedTestCaseResult } from "../config/config.types";
+
+export interface TestCaseResults {
+  comparisonToBaseReplay: DetailedTestCaseResult;
+  comparisonsToFirstHeadReplay: DetailedTestCaseResult[];
+}
+
+export const mergeResults = ({
+  comparisonToBaseReplay,
+  comparisonsToFirstHeadReplay,
+}: TestCaseResults): DetailedTestCaseResult => {
+  if (comparisonsToFirstHeadReplay.length === 0) {
+    return comparisonToBaseReplay;
+  }
+
+  // The other results all compare to the first result, so if any of them
+  // show a diff on the screenshots that originally failed then we have a flake
+  // on our hands
+  const retriedScreenshotsById: Record<
+    ScreenshotIdentifierHash,
+    ScreenshotDiffResult[]
+  > = {};
+  comparisonsToFirstHeadReplay.forEach((otherResult) => {
+    otherResult.screenshotDiffResults.forEach((result) => {
+      const hash = hashScreenshotIdentifier(result.identifier);
+      retriedScreenshotsById[hash] = retriedScreenshotsById[hash] ?? [];
+      retriedScreenshotsById[hash].push(result);
+    });
+  });
+  const newScreenshotDiffResults =
+    comparisonToBaseReplay.screenshotDiffResults.map(
+      (diffToBaseScreenshot): ScreenshotDiffResult => {
+        if (diffToBaseScreenshot.outcome === "no-diff") {
+          return diffToBaseScreenshot;
+        }
+        const hash = hashScreenshotIdentifier(diffToBaseScreenshot.identifier);
+        const diffsToFirstScreenshot = retriedScreenshotsById[hash] ?? [];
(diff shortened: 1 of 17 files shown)


packages/cli/src/parallel-tests/run-all-tests.ts
```diff
@@ -213,59 +220,22 @@ export const runAllTests = async ({
...
-      return results;
-    }
-
-    const results: DetailedTestCaseResult[] = [];
-    const progress: TestRunProgress = {
-      runningTestCases: testsToRun.length,
-      failedTestCases: 0,
-      passedTestCases: 0,
-    };
-    for (const testCase of testsToRun) {
```
Que3216, line 242: I deleted the code to run in serial, since it's essentially the same as running in parallel with parallelTasks equal to 1, and the fewer code paths to support/environmental differences the better.

packages/cli/src/parallel-tests/merge-test-results.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+  );
+
+  const noLongerHasFailures = newScreenshotDiffResults.every(
+    ({ outcome }) => outcome === "flake" || outcome === "no-diff"
+  );
+
+  return {
+    ...currentResult,
+    result:
```
Que3216, line 100: TODO: Add jsdoc on TestCase to describe what a 'flaking' test case is

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -303,3 +292,39 @@ const logRetrySummary = (
...
+    // In this case newResult is a retry of the head replay, but it failed fatally (not just a failed diff, but failed to even run)
+    // So we just ignore this retry
+    return currentMergedResult;
+  } else {
+    return mergeResults({
+      currentResult: currentMergedResult,
+      comparisonToHeadReplay: newResult,
+    });
+  }
+};
```
leafty, line 330: Style: avoid `else` clauses and nested `if` clauses.

```suggestion
  if (currentMergedResult == null && newResult == null) {
    // This means our original head replay failed fatally (not just a failed diff, but failed to even run)
    // In this case we just return it as result = fail, with no screenshot diffs
    return {
      ...testCase,
      headReplayId: "",
      result: "fail",
      screenshotDiffResults: [],
    };
  }
  
  if (currentMergedResult == null) {
    // In this case the newResult is our first head replay, our first result,
    // so lets initialize the mergedResult to this
    return newResult;
  }
  
  // If currentMergedResult is not null then newResult is a retry, containing comparison screenshots to our original head replay
  if (newResult == null) {
    // In this case newResult is a retry of the head replay, but it failed fatally (not just a failed diff, but failed to even run)
    // So we just ignore this retry
    return currentMergedResult;
  }
  
  return mergeResults({
    currentResult: currentMergedResult,
    comparisonToHeadReplay: newResult,
  });
};
```

leafty (approved):

PR #189, opened 2022-11-30
Add locking for downloads & zip extraction

Que3216:
Currently if two processes download the same file at the same time it can cause issues. For example when running with retries I'll often see process A download the zip of a replay, then process B start downloading it, but by the time process B finishes downloading it and starts extracting it, process A has already extracted the zip and deleted it.

By using lock files we also avoid two processes making the same download at the same time. Instead the 2nd will just wait for the first to finish, and then use the already downloaded file.

This uses proper-lockfile under the hood which handles things like left over lockfiles when the application crashes (it bumps mtime every 5s while lockfile is still being used, and ignores lockfiles which haven't been updated for > 10s).

### DIFF
--- a/packages/cli/src/local-data/local-data.utils.ts
+++ b/packages/cli/src/local-data/local-data.utils.ts
@@ -1,4 +1,5 @@
-import { dirname, join } from "path";
+import { mkdir, rmdir, access } from "fs/promises";
+import { join } from "path";
 import { lock, LockOptions } from "proper-lockfile";
 
 export const sanitizeFilename: (filename: string) => string = (filename) => {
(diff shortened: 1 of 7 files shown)


packages/cli/src/local-data/replays.ts
```diff
@@ -19,26 +24,32 @@ export const getOrFetchReplay: (
   const replayDir = getReplayDir(replayId);
   await mkdir(replayDir, { recursive: true });
   const replayFile = join(replayDir, `${replayId}.json`);
+  const releaseLock = await waitToAcquireLockOnFile(replayFile);
+
+  try {
```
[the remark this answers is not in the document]
Que3216: You mean the content of the try block as a function? So:

```ts
export const getOrFetchReplay: (
  client: AxiosInstance,
  replayId: string
) => Promise<any> = async (client, replayId) => {
  const replayDir = getReplayDir(replayId);
  await mkdir(replayDir, { recursive: true });
  const replayFile = join(replayDir, `${replayId}.json`);
  const releaseLock = await waitToAcquireLockOnFile(replayFile);

  try {
     return await getOrFetchReplayToFile(client, replayId, replayFile);
  } finally {
      releaseLock();
   }
}
```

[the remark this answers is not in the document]
Que3216: Or do you mean:

```ts
export const getOrFetchReplay: (
  client: AxiosInstance,
  replayId: string
) => Promise<any> = async (client, replayId) => {
  const replayDir = getReplayDir(replayId);
  await mkdir(replayDir, { recursive: true });
  const replayFile = join(replayDir, `${replayId}.json`);
  
  return withLockOnFile(replayFile, () => {
       ...
  });
}
```

packages/cli/src/local-data/local-data.utils.ts
```diff
@@ -1,3 +1,122 @@
...
+  access(filePath)
+    .then(() => true)
+    .catch(() => false);
+
+export const waitToAcquireLockOnDirectory = (
+  directoryPath: string
+): Promise<ReleaseLock> => {
+  return lock(directoryPath, {
+    retries: LOCK_RETRY_OPTIONS,
+    lockfilePath: join(directoryPath, "dir.lock"),
```
leafty, line 106: Not sure why the default is not suitable?

```suggestion
```

packages/cli/src/local-data/local-data.utils.ts
```diff
@@ -1,3 +1,122 @@
...
+const ONE_SECOND_IN_MS = 1_000;
+
+const LOCK_RETRY_OPTIONS: LockOptions["retries"] = {
+  // We want to keep on retrying till we get the maxRetryTime, so we set retries, which is a maximum, to a high value
+  retries: 1000,
+  factor: 1.05,
+  minTimeout: 500,
+  maxTimeout: 2000,
+  // Wait a maximum of 120s for the other process to finish downloading and/or extracting
+  maxRetryTime: 120 * ONE_SECOND_IN_MS,
```
leafty, line 119: nit: I think it would be nicer to use `luxon`

packages/cli/src/local-data/local-data.utils.ts
```diff
@@ -1,3 +1,122 @@
...
+    const releaseLock = await lock(lockDirectory, {
+      retries: LOCK_RETRY_OPTIONS,
+      lockfilePath: `${filePath}.lock`,
+    });
+    return async () => {
+      // Clean up our directory _before_ releasing the lock
+      // Note: it may have already been deleted: if we're just coming out of
+      // a lock released by someone else then they will have already deleted
+      // the directory
+      await rm(lockDirectory, { recursive: true, force: true });
```
leafty, line 87: note: I'm worried doing `rm` will create unnecessary I/O here, while the empty folder could be left there.

packages/cli/src/local-data/local-data.utils.ts
```diff
@@ -1,3 +1,122 @@
...
+      await releaseLock();
+    };
+  } catch (err) {
+    await rmdir(lockDirectory);
```
leafty, line 91: Also use `rm` here?

leafty (approved):

packages/cli/src/local-data/local-data.utils.ts
Que3216: This is what they recommend doing in their docs for directories (I guess it means if directory gets deleted so does the lock file). The default is designed for files.

leafty: It does not read as a recommendation to me? The default behaviour works just fine for folders.

PR #190, opened 2022-12-01
Better cli error handling

Que3216:
Currently when the code throws an Error the error is never logged out (the code in index.ts didn't seem to be working), so the command failed with no feedback as to why. This updates it to log out the error. I also updated to stop it from posting the big block of help text which can make it hard to find and read the error. Note that if you pass a command or argument that doesn't exist, miss a required arg, pass the wrong type to an arg etc. then it'll still show the help text. But not if the command handler itself throws an arbitrary error.

Side question:
 - Do we want to use `throw new Error(...)` or `process.exit` as our standard pattern? Or a mix? Note that `throw new Error(...)` means (1) it's logged to Sentry (may or may not be desired for different error types), (2) it's will execute any finally calls or catch handling, not just in the direct function the error is thrown, but in any parent function that calls that function. Normally this is desired. Maybe we lean towards `throw new Error(...)` and if we don't want all Errors to be logged to Sentry, we introduce some special types of errors like `InvalidCliArgumentsError` that doesn't get Sentry logged (or logged differently) - i.e. seperate out errors likely due to user error vs due to bug in our code?

### DIFF
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -27,7 +27,7 @@ export const main: () => void = async () => {
   const meticulousVersion = await getMeticulousVersion();
   initSentry(meticulousVersion);
 
-  const promise = yargs
+  yargs
     .scriptName("meticulous")
     .usage(
       `$0 <command>
@@ -64,13 +64,6 @@ export const main: () => void = async () => {
       (argv) => handleDataDir(argv.dataDir),
       (argv) => setOptions(argv),
     ]).argv;
-
-  if (promise instanceof Promise) {
-    promise.catch((error) => {
-      console.error(error);
-      process.exit(1);
-    });
-  }
 };
 
 main();
--- a/packages/cli/src/utils/sentry.utils.ts
+++ b/packages/cli/src/utils/sentry.utils.ts
@@ -1,5 +1,7 @@
+import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
 import * as Sentry from "@sentry/node";
 import { addExtensionMethods } from "@sentry/tracing";
+import log from "loglevel";
 import { Duration } from "luxon";
 
 const SENTRY_DSN =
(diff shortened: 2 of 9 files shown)


packages/cli/src/main.ts
```diff
@@ -64,13 +64,6 @@ export const main: () => void = async () => {
       (argv) => handleDataDir(argv.dataDir),
       (argv) => setOptions(argv),
     ]).argv;
-
-  if (promise instanceof Promise) {
-    promise.catch((error) => {
```
Que3216, line 69: This didn't seem to actually do anything as far as I could tell

packages/cli/src/utils/sentry.utils.ts
```diff
@@ -49,21 +44,31 @@ export const wrapHandler = function wrapHandler_<T>(
...
   };
 };
+
+const reportHandlerError: (error: unknown) => Promise<void> = async (error) => {
+  const logger = log.getLogger(METICULOUS_LOGGER_NAME);
+  logger.info("");
+  logger.error(error instanceof Error ? error.message : error);
+  logger.info("");
+  logger.info(
+    "Tip: run `meticulous help <command>` for help on a particular command, or `meticulous help` for a list of the available commands."
```
Que3216, line 70: Previously it'd post the full help message which could be quite long and made it harder to spot the error.

Que3216, line 70:
```suggestion
    "Tip: run `meticulous <command> --help` for help on a particular command, or `meticulous --help` for a list of the available commands."
```

packages/cli/src/main.ts
leafty: I think I had typing issues with this before. I think it's good to remove it if it works without.

leafty (approved):
LGTM :raised_hands:

packages/cli/src/main.ts
Que3216: I tested a couple of cases (error thrown in handler, error parsing args or command), and as far as I can tell it has no effect.

PR #191, opened 2022-12-01
Improve replay diff typings

Que3216:
No functional changes. These changes just make the typings far easier to use, and a little more precise. In particular Typescript doesn't like it when you do an `Omit<>` over a discriminated type union -- you lose all type narrowing -- when you filter on an outcome it doesn't filter down the type.

### DIFF
--- a/packages/api/src/replay/replay-diff.types.ts
+++ b/packages/api/src/replay/replay-diff.types.ts
@@ -36,16 +36,16 @@ export interface ScreenshotDiffOptions {
   diffPixelThreshold: number;
 }
 
-/** Represents the result of comparing two screenshots */
-export type ScreenshotDiffResult = {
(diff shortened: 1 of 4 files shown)


packages/api/src/replay/replay-diff.types.ts
```diff
@@ -115,7 +119,7 @@ export interface ScreenshotDiffResultFlake {
   /**
    * The original diff. Can be any outcome but for 'no-diff'.
    */
-  diffToBaseScreenshot: Omit<ScreenshotDiffResult, "identifier">;
+  diffToBaseScreenshot: SingleTryScreenshotDiffResult;
```
Que3216, line 122: Can no longer have a flake of a flake

leafty (approved):

PR #197, opened 2022-12-05
Do not delete .lock-target directories

Que3216:
Keeping the directories around was causing errors, since a second process would delete the lock-target in-between when a first process created it and started trying to acquire the lock:

```
[Error: ENOENT: no such file or directory, lstat '/github/home/.meticulous/sessions/<redacted>.json.lock-target'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '/github/home/.meticulous/sessions/<redacted>.json.lock-target'
}
```

This should fix these errors. If we find the .lock-target files hanging around distracting then we could later switch to a different technique.

### DIFF
--- a/packages/cli/src/local-data/local-data.utils.ts
+++ b/packages/cli/src/local-data/local-data.utils.ts
@@ -72,26 +72,18 @@ const waitToAcquireLockOnFile = async (
   // as something that's unique to the file, and gives the same path for a given file everytime. So we create our
   // own lock-target directory for this purpose (directory not file since mkdir is guaranteed to be synchronous).
   // The path needs to actually exist, since proper-lockfile resolves symlinks on it.
+  //
+  // Note: we don't delete the lock directory afterwards because doing so without creating race-conditions is tricky
   const lockDirectory = `${filePath}.lock-target`;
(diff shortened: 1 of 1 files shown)


packages/cli/src/local-data/local-data.utils.ts
```diff
@@ -72,26 +72,18 @@ const waitToAcquireLockOnFile = async (
...
-      await rm(lockDirectory, { recursive: true, force: true });
-      await releaseLock();
-    };
-  } catch (err) {
-    await rm(lockDirectory, { recursive: true, force: true });
-    throw err;
-  }
+  const releaseLock = await lock(lockDirectory, {
+    retries: LOCK_RETRY_OPTIONS,
+    lockfilePath: `${filePath}.lock`,
```
leafty, line 82: Note: I think the path is the same as the default here.

leafty (approved):

packages/cli/src/local-data/local-data.utils.ts
Que3216: The default would be `${filePath}.lock-target.lock`

PR #199, opened 2022-12-06
More bug fixes for run all tests algorithm

Que3216:
Fixes 3 bugs and adds unit tests (which I probably should have added in the first place). See comments on code for details.

### DIFF
--- a/packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
+++ b/packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
@@ -1,11 +1,14 @@
-import {
-  ScreenshotDiffResult,
-  ScreenshotIdentifier,
-  ScreenshotDiffResultMissingBaseAndHead,
-  SingleTryScreenshotDiffResult,
-} from "@alwaysmeticulous/api";
-import { DetailedTestCaseResult } from "../../config/config.types";
 import { mergeResults } from "../merge-test-results";
+import {
+  diff,
+  differentSize,
+  flake,
+  missingBase,
+  missingBaseAndHead,
+  missingHead,
+  noDiff,
+  testResult,
+} from "./mock-test-results";
 
 describe("mergeResults", () => {
   it("keeps the result as a failure when all retried screenshots are 'identical'", () => {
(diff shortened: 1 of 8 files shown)


packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
```diff
@@ -119,130 +122,3 @@ describe("mergeResults", () => {
     );
   });
 });
-
```
Que3216, line 122: These are just moved out into another file

packages/cli/src/parallel-tests/__tests__/mock-test-results.ts
```diff
@@ -0,0 +1,137 @@
+import {
+  ScreenshotDiffResult,
+  ScreenshotIdentifier,
+  ScreenshotDiffResultMissingBaseAndHead,
+  SingleTryScreenshotDiffResult,
+  TestCase,
+} from "@alwaysmeticulous/api";
+import { DetailedTestCaseResult } from "../../config/config.types";
+
```
Que3216, line 9: Unchanged, just pulled out

packages/cli/src/parallel-tests/execute-test-in-child-process.ts
```diff
@@ -0,0 +1,49 @@
...
+import { InitMessage, ResultMessage } from "./messages.types";
+
+export const executeTestInChildProcess = (
+  initMessage: InitMessage
+): Promise<DetailedTestCaseResult> => {
+  const logger = log.getLogger(METICULOUS_LOGGER_NAME);
+
+  const taskHandler = join(__dirname, "task.handler.js");
+
+  const deferredResult = defer<DetailedTestCaseResult>();
```
Que3216, line 15: This code is unchanged, just pulled out

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -102,75 +72,15 @@ export const runAllTestsInParallel: (
...
-      },
-    };
-    child.send(initMessage);
 
     // Handle task completion
-    deferredResult.promise
+    executeTest(testCase, isRetry)
       .catch(() => null)
       .then(async (result) => {
-        --inProgress;
```
Que3216, line 172: Bug #1: We were decrementing this too early, which meant a 2nd promise could call allTasksDone.resolve() while we're still saving results for the 1st promise

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -245,6 +152,18 @@ export const runAllTestsInParallel: (
...
+        // We only decrement inProgress at the very end,
+        // otherwise another promise may call allTasksDone.resolve() while
+        // we're still saving results
+        --inProgress;
+
+        if (queue.length === 0 && inProgress === 0) {
+          allTasksDone.resolve();
+        }
+
+        process.nextTick(checkNextTask);
```
Que3216, line 166: Bug #2: If there was a failure inside onTestFinished then we would fail without continuing to check for the next task

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -207,35 +123,26 @@ export const runAllTestsInParallel: (
...
-          --progress.runningTestCases;
-          progress.failedTestCases += mergedResult.result === "fail" ? 1 : 0;
-          progress.flakedTestCases += mergedResult.result === "flake" ? 1 : 0;
-          progress.passedTestCases += mergedResult.result === "pass" ? 1 : 0;
-          await onTestFinished?.(progress, finalResults).then(() => {
-            if (queue.length === 0 && inProgress === 0) {
-              allTasksDone.resolve();
-            }
-          });
+          progress = updateProgress(progress, mergedResult.result);
```
Que3216, line 143: Previously I was mutating progress, which didn't cause any issues, but is generally bad since it could give confusing results if e.g. a onTestFinished call back waited for some promise before reading from the progress object (may get a different result to reading from it before waiting for the promise)

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -207,35 +123,26 @@ export const runAllTestsInParallel: (
...
 
-        resultsByTestId.set(id, {
-          currentResult: mergedResult,
-          numberOfRetriesExecuted,
-        });
-
         if (isFinalResult) {
           // Cancel any replays that are still scheduled
-          queue = queue.filter(({ id }) => id !== id);
+          queue = queue.filter((otherReplay) => otherReplay.id !== id);
```
Que3216, line 140: Bug 3: Previously was comparing id to id! And so filtering everything out of the queue 🤦 .

packages/cli/src/parallel-tests/parallel-tests.handler.ts
```diff
@@ -1,48 +1,26 @@
...
-
-  appUrl: string | null;
-  useAssetsSnapshottedInBaseSimulation: boolean;
   parallelTasks: number | null;
-  deflake: boolean;
-  replayEventsDependencies: ReplayEventsDependencies;
-
   maxRetriesOnFailure: number;
 
+  executeTest: (
```
Que3216, line 14: Refactored this to just take an executeTest method. This allows the complex logic here to be unit tested without launching new node processes from the tests, mock out all the ReplayExecutionOptions etc. It also provides a nicer seperation of concerns -- splitting out the logic that deals with launching child processes and communicating with them, vs the logic that chooses when to execute new tests, how many to execute in parallel at a time, how to auto-retry on failure etc.

PR #203, opened 2022-12-07
Add before user event callback to API

Que3216:


packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts
```diff
@@ -12,8 +15,15 @@ export interface ReplayUserInteractionsOptions {
...
-  onTimelineEvent: OnReplayTimelineEventFn;
   logLevel: LogLevelDesc;
+
+  onTimelineEvent: OnReplayTimelineEventFn;
+
+  /**
+   * If present then will be called before executing each next user event,
+   * and will wait for the completion of the returned promise before continuing.
+   */
+  onBeforeUserEvent?: (options: BeforeUserEventOptions) => Promise<void>;
```
leafty, line 26: Maybe you can add:

```ts
export type OnBeforeUserEventFn = (
  options: BeforeUserEventOptions
) => Promise<void>;
```

leafty (approved):

PR #205, opened 2022-12-07
Update replay debugger to work with virtual time

Que3216:
It now works with or without skip pauses.

[Chromium - Meticulous Debugger - 7 December 2022 - Watch Video]
(https://www.loom.com/share/3d940c173fa14f229ee2c92e7040cab0)

Now that replay has grown a lot more complex re-using the replay code saves us some duplication:

image


packages/replay-debugger-ui/src/components/replay-debugger/replay-debugger.tsx
```diff
@@ -51,62 +56,12 @@ const ReplayDebuggerWrapped: FunctionComponent = () => {
...
-                "font-medium",
-                "rounded-md",
-                "shadow-sm",
-                "text-white",
-                "bg-violet-600",
-                "hover:bg-violet-700",
-                "focus:outline-none",
-                "focus:ring-2",
-                "focus:ring-offset-2",
-                "focus:ring-violet-500"
```
Que3216, line 103: I find it hard to quickly skim to get the core content/structure of the code when long lists of classes mixed in with functional structure & behaviour, or tell that these buttons/classes are identical at a glance, so pulled out VioletButton below.

packages/replayer/src/debugger/replay-debugger.ui.ts
```diff
@@ -0,0 +1,175 @@
...
+  const onAdvanceToIndex = async (newTargetIndex: number) => {
+    if (newTargetIndex <= state.index) {
+      return; // Do nothing
+    }
+    targetIndex = newTargetIndex;
+    await setState({ loading: true });
+    advanceToNextEvent?.();
+  };
+
+  const findEventTarget = async (event: ReplayableEvent) => {
```
Que3216, line 97: This and onCheckNextTarget are unchanged, just moved over from the old replay-debugger.ui.ts file

packages/replayer/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@alwaysmeticulous/common": "^2.22.0",
+    "@alwaysmeticulous/replay-debugger-ui": "^2.5.0",
```
leafty, line 22: This makes standard replay depend on a lot more packages. Can we avoid it? I think it's fine to fail with `--debugger` if the package is not installed.

packages/replayer/src/debugger/replay-debugger.ui.ts
```diff
@@ -0,0 +1,175 @@
...
+        return onPlayNextEvent();
+      }
+      if (eventType === "set-index") {
+        return onAdvanceToIndex(data.index || 0);
+      }
+      console.log(`Warning: received unknown event "${eventType}"`);
+    }
+  );
+
+  const url = "http://localhost:3005/";
```
leafty, line 150: Note: this could have available port finding logic in the future.

packages/replayer/package.json
Que3216: We only use startServer from this package so I could peerDependency it and do a require if and only if running with '--debugger' flag:

```ts
  let startServer: StartServerFn;

  try {
    const replayDebugger = await require("@alwaysmeticulous/replay-debugger");
    startServer = replayDebugger.startServer;
  } catch (error) {
    logger.error("Error: could not import @alwaysmeticulous/replay-debugger-ui");
    logger.error(error);
    logger.error("");
    logger.error("Please make sure you've installed or added a dependency on '@alwaysmeticulous/replay-debugger-ui'. It is not installed automatically, and is required to use the '--debugger' flag.");
    process.exit(1);
  }
```

leafty: Yes, you can lookup the `peerDependencies` in `cli` and see how to mark it as optional.

leafty (approved):
LGTM

Nice to see this update :+1: :tada:

PR #208, opened 2022-12-14
Add support for session selection from the Meticulous API

leafty:
This change pulls the project configuration from the Meticulous API and merge the test cases to run with the ones present in the `meticulous.json` file (if it exists).

[SIM-386](https://toil.kitemaker.co/qpgVHi-Meticulous/pPGwrM-Engineering/items/386)

### DIFF
--- a/packages/cli/src/utils/run-all-tests.utils.ts
+++ b/packages/cli/src/utils/run-all-tests.utils.ts
@@ -8,9 +8,6 @@ import { DetailedTestCaseResult } from "../config/config.types";
 export const mergeTestCases = (
   ...testSuites: (TestCase[] | null | undefined)[]
 ): TestCase[] => {
-  const logger = log.getLogger(METICULOUS_LOGGER_NAME);
-  logger.debug(testSuites.length);
-
   const seenSessionIds = new Set<string>();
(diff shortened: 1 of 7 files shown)


alexivanov (approved):

packages/cli/src/utils/run-all-tests.utils.ts
```diff
@@ -5,6 +5,25 @@ import log from "loglevel";
...
+    if (testSuite == null) {
+      return [];
+    }
+    return testSuite.flatMap((testCase) => {
+      if (seenSessionIds.has(testCase.sessionId)) {
+        return [];
+      }
+      seenSessionIds.add(testCase.sessionId);
+      return [testCase];
+    });
```
Que3216, line 23: Note if we want the repeat stuff that the main meticulous repo does to still work then we may need to do:

```ts
    const novelTestCases = testSuite.filter((testCase) => !seenSessionIds.has(testCase.sessionId));
    testSuite.forEach((testCase) => seenSessionIds.add(testCase.sessionId));
    return novelTestCases;
```

So that it only dedups across testSuites and not within testSuites.

PR #217, opened 2023-01-11
handle simulation timeout errors

alexivanov:
This terminates the replay early if timeout occurs while replaying user interaction events

### DIFF
--- a/packages/replayer/src/replayer.ts
+++ b/packages/replayer/src/replayer.ts
@@ -293,7 +293,11 @@ export const replayEvents: ReplayEventsFn = async (options) => {
       }
     }
   } catch (error) {
-    logger.error("Error thrown during replay", error);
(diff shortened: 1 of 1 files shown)


leafty (approved):

packages/replayer/src/replayer.ts
```diff
@@ -293,7 +293,11 @@ export const replayEvents: ReplayEventsFn = async (options) => {
       }
     }
   } catch (error) {
-    logger.error("Error thrown during replay", error);
```
Que3216, line 296: Maybe still worth logging the error, particularly if it's not a MeticulousTimeoutError? I know it goes to the timeline, but if debugging an issue you may not think to look at the timeline in addition to the terminal logs.

alexivanov: Oops, removal of this wasn't intended. reverted in #223

PR #240, opened 2023-01-16
Expose Sentry initialization and make sample rate configurable

alexivanov:
The GitHub action calls the replay handler directly, skipping the init logic within the CLI.

This PR exposes the Sentry initialisation logic to enable recording GitHub action runs performance data.

Additionally this makes the sample rate configurable and handles cases where replay handler is called within a Sentry transaction (so action runs can be grouped in a single transaction)

### DIFF
--- a/packages/cli/src/commands/replay/replay.command.ts
+++ b/packages/cli/src/commands/replay/replay.command.ts
@@ -99,11 +99,19 @@ export const replayCommandHandler = async ({
     );
   }
 
-  const transaction = Sentry.startTransaction({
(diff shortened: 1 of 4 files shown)


Que3216 (approved):

Que3216:
The plan is to set 'METICULOUS_TELEMETRY_SAMPLE_RATE' to '0.01' in report-diffs-action?

alexivanov:
>The plan is to set 'METICULOUS_TELEMETRY_SAMPLE_RATE' to '0.01' in report-diffs-action?

Yep, probably with around 10% sample rate though given that's on per-test run basis.

PR #244, opened 2023-01-19
Allow skipping upload stage when replaying

Que3216:
For many sessions uploading the archive takes up to 9 times as long as the replay itself, so being able to skip the upload stage can speed up testing (e.g. when running command repeatedly in a bash for loop to check for flakey issues).

### DIFF
--- a/packages/cli/src/commands/replay/utils/exit-early-if-skip-upload-env-var-set.ts
+++ b/packages/cli/src/commands/replay/utils/exit-early-if-skip-upload-env-var-set.ts
@@ -1,0 +1,32 @@
+import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
+import log from "loglevel";
+
+const METICULOUS_SKIP_UPLOAD_ENV_VAR = "METICULOUS_SKIP_UPLOAD";
+
+// Uploading the zip can take a long time, so we expose a env variable to support skipping it
+// We don't expose this as a first-class CLI option because we don't want to pollute the list of CLI options
+export const exitEarlyIfSkipUploadEnvVarSet = (
+  baseReplayId: string | null | undefined
+) => {
+  if (!shouldSkipUpload()) {
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/replay/utils/exit-early-if-skip-upload-env-var-set.ts
```diff
@@ -0,0 +1,32 @@
...
+// Uploading the zip can take a long time, so we expose a env variable to support skipping it
+// We don't expose this as a first-class CLI option because we don't want to pollute the list of CLI options
+export const exitEarlyIfSkipUploadEnvVarSet = (
+  baseReplayId: string | null | undefined
+) => {
+  if (!shouldSkipUpload()) {
+    return;
+  }
+
+  if (baseReplayId != null) {
```
alexivanov, line 15: technically throwing will still have the same outcome of exiting early, albeit with a different exit code. shall we handle this upstream / log a warning/error instead?

alexivanov (approved):

packages/cli/src/commands/replay/utils/exit-early-if-skip-upload-env-var-set.ts
Que3216: In wrapHandler we `process.exit(1)` and log the error on an error: https://github.com/alwaysmeticulous/meticulous-sdk/blob/69d9c8e364465c048c92f81dade512cf5046fb4b/packages/cli/src/utils/sentry.utils.ts#L59

I was thinking to try and move towards using `throw new Error` instead of `process.exit` where we can because sometimes these utilities end up getting re-used as imports called by other modules (either directly or transitively), and a hidden process.exit can thwart `try/catch`s etc. In this case it'd probably be fine, but if we get to a consistent pattern of just using throw new Error and relying on wrapHandler to handle it correctly then we don't need to think each time, we just always throw new Error, and always know it's the right thing.

PR #246, opened 2023-01-24
Fix run-all-tests sentry

alexivanov:
This fixes Sentry when running simulations within a child process,
such as when running via `run-all-tests` or the Github Action.


leafty (approved):
LGTM

PR #295, opened 2023-02-08
allow setting a user agent override for simulations

alexivanov:


leafty (approved):
LGTM

Is there a specific reason for this? Custom user agents tend to be blocked by things like Cloudflare (which could provide extra replay protection I guess).

alexivanov:
>Is there a specific reason for this? Custom user agents tend to be blocked by things like Cloudflare (which could provide extra replay protection I guess).

Mostly for the (rare) cases where the inverse is true, i.e where we need to provide a specific user agent to access assets during simulation.

PR #311, opened 2023-02-27
Special handing for NextJS start URLs during simulation

alexivanov:
### Changes

This PR adds special handling for simulation of sessions with SSRed Next JS pages. We want to avoid hitting SSR pages and instead use Next's router logic of . To achieve this we start simulation from a known non-SSR page, currently `_error` which always 404s and then navigate to the correct start URL via `window.next.router.replace()`. The network stubbing snippet now handles successful mocking of the initial `_next/data` request/

### DIFF
--- a/packages/replayer/src/next/utils.ts
+++ b/packages/replayer/src/next/utils.ts
@@ -1,0 +1,11 @@
+import { SessionData } from "@alwaysmeticulous/api";
+
+export const isSSRPage: (sessionData: SessionData) => boolean = (sessionData) =>
+  !!sessionData.applicationSpecificData?.nextJs?.props?.__N_SSP;
+
+export const getNextJs404PageUrl: (startUrl: string) => string = (startUrl) => {
+  const url = new URL(startUrl);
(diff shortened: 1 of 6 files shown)


packages/replayer/src/next/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { SessionData } from "@alwaysmeticulous/api";
+
+export const isSSRPage: (sessionData: SessionData) => boolean = (sessionData) =>
+  !!sessionData.applicationSpecificData?.nextJs?.props?.__N_SSP;
+
+export const getNextJs404PageUrl: (startUrl: string) => string = (startUrl) => {
+  const url = new URL(startUrl);
+  // _error will always be a 404 page in Next.JS. See https://nextjs.org/docs/advanced-features/custom-error-page#caveats
+  url.pathname = "/_error";
```
Que3216, line 9: Such a wonderful hack

Que3216 (approved):
Looks great!

PR #330, opened 2023-03-20
Compare to multiple base replay ids

Que3216:
This shouldn't be merged until the corresponding change to our backend is merged, but putting it up for review.

### DIFF
--- a/packages/cli/src/commands/replay/utils/compute-diff.ts
+++ b/packages/cli/src/commands/replay/utils/compute-diff.ts
@@ -1,12 +1,11 @@
-import { mkdir } from "fs/promises";
-import { basename, join } from "path";
+import { mkdir, opendir, rm, rmdir } from "fs/promises";
+import { join } from "path";
 import {
   ScreenshotAssertionsEnabledOptions,
   ScreenshotDiffResult,
   ScreenshotIdentifier,
 } from "@alwaysmeticulous/api";
 import { AxiosInstance } from "axios";
-import stringify from "fast-json-stable-stringify";
 import log from "loglevel";
 import { downloadFile } from "../../../api/download";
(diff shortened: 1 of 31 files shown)


packages/cli/src/commands/screenshot-diff/screenshot-diff.command.ts
```diff
@@ -305,27 +249,30 @@ const handler: (options: Options) => Promise<void> = async ({
...
 
+  const baseReplayScreenshots = await getScreenshotFiles(baseScreenshotsDir);
+  const headReplayScreenshots = await getScreenshotFiles(headScreenshotsDir);
+
   const results = await diffScreenshots({
     client,
     baseReplayId,
     headReplayId,
     baseScreenshotsDir,
     headScreenshotsDir,
```
alexivanov, line 260: Given we're now passing the full list of screenshots can the `(base|head)ScreenshotsDir` be dropped as params?

packages/cli/src/commands/replay/utils/compute-diff.ts
```diff
@@ -27,42 +33,128 @@ export interface ComputeAndSaveDiffOptions {
...
+  screenshots: ScreenshotLocator[],
+  directory: string
+) => {
+  // Download in batches of 20
+  const screenshotBatches = chunk(screenshots, 20);
+  for (const screenshotBatch of screenshotBatches) {
+    const screenshotDownloads = screenshotBatch.map((screenshot) => {
+      const filePath = join(
+        directory,
+        basename(new URL(screenshot.screenshotUrl).pathname)
```
alexivanov, line 131: Should we use the screenshot identifier here to generate a file name? I don't think we should assume that / rely on screenshot URL paths matching the identifier format.

alexivanov (approved):

PR #332, opened 2023-03-23
Make replay-next-node responsible for taking end state screenshots

Que3216:
Note: this PR must merge _after_ the corresponding replay-next-node PR.


packages/sdk-bundles-api/src/replay/sdk-to-bundle/index.ts
```diff
@@ -7,16 +7,25 @@ import {
...
 
 export interface ReplayUserInteractionsOptions {
+  /**
+   * A semantic version number for the SDK calling into the replay code.
+   *
+   * This version number is bumped on every API change, and allows the replay
+   * code to detect if it's being called by an old version, and if so throw
+   * and request the user updates to a newer version.
+   */
+  sdkSemanticVersion: number;
```
alexivanov, line 17: Nice 💯

packages/replayer/src/replay.utils.ts
```diff
@@ -209,7 +209,7 @@ export const getSessionDuration: (
     userEventEventLog[userEventEventLog.length - 1]?.timeStamp;
 
   if (maxUserEventTimestamp == null) {
```
alexivanov, line 211: Is the session duration based on `rrweb` still accurate given we now selectively record rrweb data for sessions?

alexivanov (approved):

packages/replayer/src/replay.utils.ts
Que3216: It handles the case of rrweb not being present. I left this in for backwards compatibility -- otherwise you may get differing screenshots for replays of sessions that do have rrweb data compared to the screenshots that used the rrweb duration instead of the last user event timestamp.

I could remove it, but customers may get some one off screenshot diffs (for both virtual time & non-virtual time). If we do remove it probably best in a seperate PR.

PR #335, opened 2023-03-27
Add flag to re-run tests multiple times

alexivanov:
### Changes

This PR adds an optional argument to `runAllTests`. As name suggests it runs tests up to N times, regardless of whether it has a difference to base on the head run.

This is useful for determining flake rate of a given session.

### DIFF
--- a/packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
+++ b/packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
@@ -123,4 +123,16 @@ describe("mergeResults", () => {
       ])
     );
   });
+
+  it("returns a fail outcome if the original outcome was a pass and there are flakes", () => {
+    const currentResult = testResult("pass", [noDiff(0), noDiff(1)]);
+    const comparisonToHeadReplay = testResult("fail", [diff(0), noDiff(1)]);
+    const mergedResult = mergeResults({
+      currentResult,
+      comparisonToHeadReplay,
+    });
+    expect(mergedResult).toEqual(
+      testResult("fail", [flake(0, noDiff(), [diff()]), noDiff(1)])
+    );
+  });
 });
--- a/packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
+++ b/packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
@@ -123,16 +123,4 @@ describe("mergeResults", () => {
       ])
     );
   });
-
-  it("returns a fail outcome if the original outcome was a pass and there are flakes", () => {
-    const currentResult = testResult("pass", [noDiff(0), noDiff(1)]);
-    const comparisonToHeadReplay = testResult("fail", [diff(0), noDiff(1)]);
-    const mergedResult = mergeResults({
-      currentResult,
-      comparisonToHeadReplay,
-    });
-    expect(mergedResult).toEqual(
-      testResult("fail", [flake(0, noDiff(), [diff()]), noDiff(1)])
-    );
-  });
 });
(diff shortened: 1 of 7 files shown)


packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
```diff
@@ -121,4 +123,16 @@ describe("mergeResults", () => {
       ])
     );
   });
+
+  it("returns a fail outcome if the original outcome was a pass and there are flakes", () => {
```
Que3216, line 127: Isn't this basically identical to the "doesn't ignore diffs to screenshots which originally passed" test case, but with the `noDiff(0), diff(1)` order swapped? Wonder if we could just delete the "doesn't ignore diffs to screenshots which originally passed" test case.

Que3216 (commented):
Added one comment, but haven't reviewed it all yet -- I'll review the rest tomorrow.

packages/cli/src/parallel-tests/__tests__/merge-test-results.spec.ts
alexivanov: oops that's true. will drop this test

packages/cli/src/parallel-tests/__tests__/parrallel-tests.handler.spec.ts
```diff
@@ -201,6 +206,41 @@ describe("runAllTestsInParallel", () => {
       ),
     ]);
   });
+
+  it("retries N times if rerunTestsNTimes > 0", async () => {
+    let rerunNum = 0;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    // const { baseTestRunId, ...testCaseWithoutBase } = testCase(0);
```
Que3216, line 213:
```suggestion
```

packages/cli/src/parallel-tests/merge-test-results.ts
```diff
@@ -111,6 +110,9 @@ export const mergeResults = ({
     result:
       currentResult.result === "fail" && noLongerHasFailures
         ? "flake"
+        : currentResult.result === "pass" &&
+          comparisonToHeadReplay.result === "fail"
+        ? "fail"
```
Que3216, line 115: Should this be flake? (I'm assuming comparisonToHeadReplay is "fail" if any screenshot is different between the head and the retry)

Side note: I find the nested ? statements a bit hard to parse, so I'll sometimes split out into a function, or use if statements:

```
const screenshotDiffResultsByBaseReplayId: groupScreenshotDiffResults(
      newScreenshotDiffResults
    );
if (currentResult.result === "fail" && noLongerHasFailures) {
   return { ...currentResult, result: "flake", screenshotDiffResultsByBaseReplayId}
} else if (...) {
   ...
}
```

Que3216: i.e. copy the logic from deflake-tests-handler:

```
  const result = hasNotableDifferences(
    Object.values(screenshotDiffResultsByBaseReplayId).flat()
  )
    ? "fail"
    : "pass";
```

Que3216: Note: it may also make sense to refactor so that we no longer calculate the aggregated result. The backend could calculate it when it ingests the result or something. Otherwise it's a bit of a pain passing it around and mantaining it. I'm not sure we even use it anymore.

alexivanov: https://github.com/alwaysmeticulous/meticulous-sdk/pull/335/commits/a3eb7bec3a2a6f2b4288ba0f5d1d0c19437d4d82 refactors the logic

Que3216 (approved):

PR #336, opened 2023-03-28
Support versioned screenshots

Que3216:


packages/api/src/replay/test-run.types.ts
```diff
@@ -23,8 +23,6 @@ export interface TestCaseReplayOptions extends Partial<ScreenshotDiffOptions> {
    */
   simulationIdForAssets?: string | undefined;
 
-  screenshotSelector?: string;
```
Que3216, line 26: We no longer use this

packages/replayer/src/screenshot.utils.ts
```diff
@@ -1,49 +0,0 @@
-import { join } from "path";
```
Que3216, line 1: We no longer use this

PR #337, opened 2023-03-28
Remove screenshot diff command

Que3216:


packages/cli/src/commands/replay/screenshot-diffing/diff-downloaded-screenshots.ts
```diff
@@ -0,0 +1,149 @@
...
+  ScreenshotDiffOptions,
+  ScreenshotDiffResult,
+} from "@alwaysmeticulous/api";
+import stringify from "fast-json-stable-stringify";
+import { compareImages } from "../../../image/diff.utils";
+import { readPng } from "../../../image/io.utils";
+import { IdentifiedScreenshotFile } from "../../../local-data/replays";
+import { writeScreenshotDiff } from "../../../local-data/screenshot-diffs";
+
+export const diffDownloadedScreenshots = async ({
```
Que3216, line 12: This function is unchanged. Just moved out from screenshot-diff.command.ts and renamed.

Gabriel-H (approved):

PR #338, opened 2023-03-28
Split out replay orchestrator [Part Complete]

Que3216:


packages/api/src/replay/test-run.types.ts
```diff
@@ -99,3 +100,27 @@ export interface TestRunGitHubWorkflowDispatchContext {
...
+
+export interface TestCaseResult extends TestCase {
+  headReplayId: string;
+
+  /**
+   * A test case is marked as a flake if there were screenshot comparison failures,
+   * but for every one of those failures regenerating the screenshot on head sometimes gave
+   * a different screenshot to the original screenshot taken on head.
+   */
+  result: "pass" | "fail" | "flake";
```
alexivanov, line 125: Might be worth extracting the result into it's own type as well.

alexivanov (approved):

PR #340, opened 2023-03-29
Expand list of test cases to run with those from the base test run

alexivanov:
This PR updates the `runAllTests` logic to expand the list of test cases to include those from the base test run. This is necessary to prevent a drop in coverage when the project config test cases are updated. Such updates have the potential to entirely replace the set of sessions to run and will occur more frequently in the future.

### DIFF
--- a/packages/replay-orchestrator/src/test-run/parallel-tests/run-all-tests.ts
+++ b/packages/replay-orchestrator/src/test-run/parallel-tests/run-all-tests.ts
@@ -19,7 +19,6 @@ import {
   ReplayExecutionOptions,
 } from "@alwaysmeticulous/common";
 import { loadReplayEventsDependencies } from "@alwaysmeticulous/download-helpers/dist/scripts/replay-assets";
-import { AxiosInstance } from "axios";
 import log from "loglevel";
 import { createReplayDiff } from "../../api/replay-diff.api";
 import {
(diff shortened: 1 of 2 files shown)


packages/replay-orchestrator/src/test-run/parallel-tests/run-all-tests.ts
```diff
@@ -402,3 +418,18 @@ const getTestCasesWithBaseTestRunId = async ({
...
 };
+
+const shouldUseBaseTestRunTestCases = async ({
+  client,
+  baseTestRunId,
+}: {
+  client: AxiosInstance;
+  baseTestRunId: string;
+}) => {
+  const testRun = await getTestRun({ client, testRunId: baseTestRunId });
```
Que3216, line 429: Haven't we already loaded this, stored as fallbackTestRun?

Que3216 (approved):
Looks good, though I don't know if we need the extra call

packages/replay-orchestrator/src/test-run/parallel-tests/run-all-tests.ts
alexivanov: 🤦 Yep, https://github.com/alwaysmeticulous/meticulous-sdk/pull/340/commits/8c90537f00c991b576784068958c6314190cc0e6 drops the extra call

packages/replay-orchestrator/src/test-run/parallel-tests/run-all-tests.ts
```diff
@@ -402,3 +418,18 @@ const getTestCasesWithBaseTestRunId = async ({
...
+
+  // We only want to include the test cases from the base test run if it has no base test run itself, i.e the
+  // the base test run is for a "push" event.
+  // This safeguards against the test run expanding indefinitely.
```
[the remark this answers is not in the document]
Que3216: (i.e. we could store `fallbackTestRun?.configData.sessionIdsExecutedForBackCompatOnly`, and then filter these ones out)

PR #345, opened 2023-03-30
Fix flake detection bug

Que3216:
### Change overview

Allows specifying either a baseReplayId (compareTo.type === "specific-replay") or a baseTestRunId (compareTo.type === "best-replay-for-session-in-test-run") to compare a replay against. We use the compareTo.type === "specific-replay" option when we're performing a retry, and set the compareTo.replayId to the replay id of the original first try. 

### Testing performed

 - [x] Live assets tests
 - [x] Comparison tests with maxRetriesOnFailure = 0 to check a case where it should give a diff, and not a flake
 - [x] Comparison tests with maxRetriesOnFailure = 3 to check a case where it should give a diff, and not a flake
 - [x] A test with a flake


packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -175,6 +186,12 @@ export const replayCommand = buildCommand("simulate")
       string: true,
       description: "Path to cookies to inject before simulation",
     },
+    baseReplayId: {
```
Que3216, line 189: I brought this back now that we have the functionality anyway.

packages/cli/src/commands/replay/replay.command.ts
```diff
@@ -175,6 +186,12 @@ export const replayCommand = buildCommand("simulate")
...
+      string: true,
+      description:
+        "Base simulation id to diff the final state screenshot against",
```
alexivanov, line 192: it this diffing only the final state screenshot?

packages/replay-orchestrator/src/replay/utils/exit-early-if-skip-upload-env-var-set.ts
```diff
@@ -1,20 +1,26 @@
...
-  baseTestRunId: string | null | undefined
+  screenshottingOptions: ScreenshotComparisonOptions
 ) => {
   if (!shouldSkipUpload()) {
     return;
   }
 
-  if (baseTestRunId != null) {
+  if (
+    screenshottingOptions.enabled === true &&
```
alexivanov, line 17:
```suggestion
    screenshottingOptions.enabled &&
```

packages/replay-orchestrator/src/replay/utils/exit-early-if-skip-upload-env-var-set.ts
```diff
@@ -1,20 +1,26 @@
...
+    screenshottingOptions.compareTo.type !== "do-not-compare"
+  ) {
     throw new Error(
-      `Cannot specify baseTestRunId and compare to base results when ${METICULOUS_SKIP_UPLOAD_ENV_VAR} is set to true`
+      `Cannot compare to base results when ${METICULOUS_SKIP_UPLOAD_ENV_VAR} is set to true. ${METICULOUS_SKIP_UPLOAD_ENV_VAR}
+      cannot be used when specifying anything to compare to (a --baseTestRunId, or --baseReplayId etc.)
```
alexivanov, line 22:
```suggestion
      cannot be used when specifying anything to compare to (a --baseTestRunId, or --baseCommitSha etc.)
```

packages/replay-orchestrator/src/replay/screenshot-diffing/download-and-diff-screenshots.ts
```diff
@@ -17,37 +25,80 @@ import { getScreenshotFilename } from "./utils/get-screenshot-filename";
...
+      sessionId,
+      headReplayDir,
+      headReplayId,
+      compareTo: screenshottingOptions.compareTo,
+      logger,
+    });
+  } else if (screenshottingOptions.compareTo.type === "specific-replay") {
+    logger.info(
+      `Diffing screenshots against replay ${screenshottingOptions.compareTo.replayId}`
+    );
```
alexivanov, line 64: looks like this is part of `diffScreenshotsAgainstReplay`
```suggestion
```

packages/replay-orchestrator/src/test-run/parallel-tests/parallel-tests.handler.ts
```diff
@@ -299,3 +297,42 @@ const updateProgress = (
...
   };
 };
+
+const createTaskToRetry = (
+  testCaseId: number,
+  testTask: TestTask,
+  firstTryReplayId: string
+): RerunnableTestTask => {
+  const { screenshottingOptions } = testTask;
+  if (screenshottingOptions.enabled === false) {
```
alexivanov, line 307:
```suggestion
  if (!screenshottingOptions.enabled) {
```

alexivanov (approved):

packages/cli/src/commands/replay/replay.command.ts
Que3216, line 192:
```suggestion
        "Base simulation id to diff the screenshots against",
```

PR #347, opened 2023-03-31
Rename replayer to replay-launcher

Que3216:


packages/downloading-helpers/src/scripts/replays.ts
```diff
@@ -51,9 +51,9 @@ export const getOrFetchReplayArchive = async (
 
   try {
     const replayArchiveFile = join(replayDir, `${replayId}.zip`);
-    const paramsFile = join(replayDir, "replayEventsParams.json");
+    const paramsFile = join(replayDir, "metadata.json");
```
Que3216, line 54: metadata.json should always exist, while I've renamed replayEventsParams.json

alexivanov (approved):

PR #349, opened 2023-04-03
maxSemanticVersion typings

Que3216:
Type as a const of the latest known version, rather than a number, to ensure that all clients bump the version number passed when they upgrade to the types.

### DIFF
--- a/packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
+++ b/packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
@@ -35,8 +35,11 @@ export interface ReplayAndStoreResultsOptions {
    * code to detect if it's being called by client that is not compatible with the latest version,
    * and if so throw an OutOfDateClientError. It is then up to the client to display a message to ask
    * the user to update to a newer version.
+   *
+   * Note: this is typed as a const of the latest known version, rather than a number, to ensure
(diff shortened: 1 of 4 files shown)


packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
```diff
@@ -35,8 +35,11 @@ export interface ReplayAndStoreResultsOptions {
    * code to detect if it's being called by client that is not compatible with the latest version,
    * and if so throw an OutOfDateClientError. It is then up to the client to display a message to ask
    * the user to update to a newer version.
+   *
+   * Note: this is typed as a const of the latest known version, rather than a number, to ensure
+   * that all clients bump the version number passed when they upgrade to the types.
    */
-  maxSemanticVersionSupported: number;
+  maxSemanticVersionSupported: 1;
```
Que3216, line 42: Workflow when making a break:
 1. Open SDK PR to bump this number, make the API changes, and update the SDK to work with both the new and old API version. Release SDK and report-diffs-action.
 2. Open main repo PR to update it to the new version of these typings, and to throw when maxSemanticVersionSupported < the new version number. Merge & release.
 3. Open SDK PR to remove support for the old version.

Que3216: Actually, I've added this to a README

PR #356, opened 2023-04-03
Use replay orchestrator launcher

Que3216:
I'll remove the replay-orchestrator and replay-launcher packages in a FLUP PR.

This PR should not merge until the PR that publishes the bundle scripts merges.


packages/replay-orchestrator/src/test-run/parallel-tests/execute-test-run.ts
```diff
@@ -259,6 +259,7 @@ export const executeTestRun = async ({
             testRunId: testRun.id,
             replayEventsDependencies,
             cookiesFile: null,
+            logLevel: logger.getLevel(),
```
Que3216, line 262: Note: this is duplicated, but I'll refactor in main repo. I'm going to delete this package in next PR

packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-test-run.ts
```diff
@@ -61,6 +62,8 @@ export interface ExecuteTestRunOptions {
 
   baseTestRunId: string | null;
 
+  logLevel: LogLevelNumbers;
```
alexivanov, line 65: Should we be more liberal here and allow `LogLevelDesc` instead?

```
    /**
     * Possible log level descriptors, may be string, lower or upper case, or number.
     */
    type LogLevelDesc = LogLevelNumbers | LogLevelNames | 'silent' | keyof LogLevel;
```

alexivanov (approved):
one small comment about the logLevel type but other than that LGTM! 🚀

PR #361, opened 2023-04-06
Add base & head refs to test run env

Que3216:


packages/api/src/sdk-bundle-api/sdk-to-bundle/test-run-environment.ts
```diff
@@ -25,8 +25,14 @@ export interface TestRunGitHubPullRequestContext {
...
 
+  /** Base ref (usually /refs/head/<branch>) */
+  baseRef?: string;
+
   /** Head commit hash */
   headSha: string;
+
+  /** Head ref (usually /refs/head/<branch>) */
+  headRef?: string;
 }
```
Que3216, line 36: Note we may also want to expose the details required to construct a URL to the test run

alexivanov (approved):

PR #381, opened 2023-04-20
add chromeExecutablePath to replay / test run options

alexivanov:


Que3216 (approved):

packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
```diff
@@ -89,8 +90,10 @@ const handler: (options: Options) => Promise<void> = async ({
 
   const parrelelTasks = parallelize ? parrelelTasks_ : 1;
   const commitSha = (await getCommitSha(commitSha_)) || "unknown";
+  const chromeExecutablePath = executablePath();
```
Que3216, line 93: Would it make sense to push this down to happen inside '@alwaysmeticulous/replay-orchestrator-launcher'? Then things like report-diffs-action, or places where we use ''@alwaysmeticulous/replay-orchestrator-launcher' inside the main Meticulous monorepo for the tests, or anywhere we use it in future, would get the correct behaviour by default?

Que3216: @alwaysmeticulous/replay-orchestrator-launcher can then have the puppeteer dependency, and so a dep on @alwaysmeticulous/replay-orchestrator-launcher triggers the Chromium download at preinstall time

Que3216 (approved):

packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
alexivanov: 👍 sounds good to me. Originally I thought about keeping the responsibility of providing a puppeteer version to the "end" client (CLI / CI Action etc) but this might not be desirable

PR #404, opened 2023-05-15
define default execution / screenshotting options

alexivanov:
Goal is to unify all call sites which share the same options


Que3216 (approved):
Nice! Can even use these in the CLI too if you like https://github.com/alwaysmeticulous/meticulous-sdk/blob/main/packages/cli/src/command-utils/common-options.ts#L33, though not essential

PR #430, opened 2023-06-26
Turn on move before click by default

Que3216:
We'll need to bump the logic version when we bump any dependency on meticulous-sdk.

### DIFF
--- a/packages/cli/src/command-utils/common-options.ts
+++ b/packages/cli/src/command-utils/common-options.ts
@@ -9,6 +9,11 @@
 //
 // This then makes the difference in behaviour explicit
 
+import {
+  DEFAULT_EXECUTION_OPTIONS,
+  DEFAULT_SCREENSHOTTING_OPTIONS,
+} from "@alwaysmeticulous/common";
+
 export const OPTIONS = {
   apiToken: {
     string: true,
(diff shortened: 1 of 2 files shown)


packages/cli/src/command-utils/common-options.ts
```diff
@@ -17,65 +22,67 @@ export const OPTIONS = {
...
     description: "Simulate mouse movement before clicking",
-    default: true,
+    default: DEFAULT_EXECUTION_OPTIONS.moveBeforeClick,
   },
   diffThreshold: {
     number: true,
     description:
       "Acceptable maximum proportion of changed pixels, between 0 and 1. If this proportion is exceeded then the test will fail.",
-    default: 0.01,
+    default: DEFAULT_SCREENSHOTTING_OPTIONS.diffOptions.diffThreshold,
```
Que3216, line 66: This changes it from 0.01 (10k pixels on a 1000 by 1000px screenshot, or 100 pixels by 100 pixels) to 0.00001 (10 pixels on a 1000 by 1000px screenshot). But I think it makes sense for the CLI defaults to match the CI defaults, otherwise debugging can be misleading. Also 100 pixels by 100 pixels was _**far**_ too large

packages/cli/src/command-utils/common-options.ts
```diff
@@ -9,6 +9,11 @@
 //
 // This then makes the difference in behaviour explicit
 
+import {
+  DEFAULT_EXECUTION_OPTIONS,
+  DEFAULT_SCREENSHOTTING_OPTIONS,
+} from "@alwaysmeticulous/common";
```
alexivanov, line 15: Nice 👍

alexivanov (approved):
🚀

PR #435, opened 2023-06-27
Replace adm-zip with extract-zip

alexivanov:
Extract-zip supports streaming which reduces the memory footprint of extracting replay zips

### DIFF
--- a/packages/downloading-helpers/src/scripts/replays.ts
+++ b/packages/downloading-helpers/src/scripts/replays.ts
@@ -5,8 +5,8 @@ import {
   getMeticulousLocalDataDir,
   METICULOUS_LOGGER_NAME,
 } from "@alwaysmeticulous/common";
-import Zip from "adm-zip";
 import { AxiosInstance } from "axios";
+import extract from "extract-zip";
 import log from "loglevel";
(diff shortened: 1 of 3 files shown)


packages/downloading-helpers/src/scripts/replays.ts
```diff
@@ -71,8 +71,7 @@ export const getOrFetchReplayArchive = async (
 
     await downloadFile(downloadUrlData.dowloadUrl, replayArchiveFile);
 
-    const zipFile = new Zip(replayArchiveFile);
-    zipFile.extractAllTo(replayDir, /*overwrite=*/ true);
+    await extract(replayArchiveFile, { dir: replayDir });
```
Que3216, line 74: I guess `join(getMeticulousLocalDataDir(), "replays", replayId)` could have had existing files. Does 'extract-zip' overwrite by default? (I couldn't tell from the README or tests

Que3216 (approved):

packages/downloading-helpers/src/scripts/replays.ts
alexivanov: yep, looks like it overrides by default

PR #439, opened 2023-06-27
API typings for GitLab MR test runs

dbook13:


packages/api/src/sdk-bundle-api/sdk-to-bundle/test-run-environment.ts
```diff
@@ -64,3 +64,9 @@ export interface TestRunGitHubWorkflowDispatchContext {
   /** Resolved head commit hash */
   headSha: string;
 }
+
+export type TestRunGitLabContext = TestRunGitLabPullRequestContext;
+export interface TestRunGitLabPullRequestContext {
+  type: "gitlab";
+  event: "pull-request";
+}
```
Que3216, line 72: We may also want the pull request title URL etc. if you want it to display in the UI on the test run page: 

image

(though I don't know if that's easy to grab from GitLab API)

Also can always add it later / this PR looks good to merge as is.

dbook13: Updated to include the fields necessary to render that component. Also included baseSha and headSha since those are both included in github pull request context.

PR #442, opened 2023-06-30
v2.67.0

dbook13:
Is this how we create publish new versions to NPM?


Que3216:
It's not -- to publish check out the main branch locally, pull to get latest, run `yarn run publish` and then choose a minor version bump (we're not being good semver'ers right now, so mainly just doing minor bumps).

Que3216:
Doing the above will push up the right tags etc., and trigger https://github.com/alwaysmeticulous/meticulous-sdk/blob/main/.github/workflows/publish.yaml#L53

dbook13:
Thanks! Looks like it worked now

PR #473, opened 2023-08-21
More screenshotting options

alexivanov:
This PR adds a couple extra screenshotting options -- optional wait before taking a screenshot and a flag to take a full page screenshot rather than viewport one.

Additionally, these will have default values on project level so `Project` API type is extended to include those.

### DIFF
--- a/packages/downloading-helpers/src/scripts/replay-assets.ts
+++ b/packages/downloading-helpers/src/scripts/replay-assets.ts
@@ -41,11 +41,11 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
     );
     const filePath = join(await getOrCreateAssetsDir(), assetFileName);
 
-    // if (entry && etag !== "" && etag === entry.etag) {
-    //   logger.debug(`${fetchUrl} already present`);
-    //   releaseLock();
-    //   return filePath;
-    // }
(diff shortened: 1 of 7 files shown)


packages/downloading-helpers/src/scripts/replay-assets.ts
```diff
@@ -41,11 +41,11 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
     );
     const filePath = join(await getOrCreateAssetsDir(), assetFileName);
 
-    if (entry && etag !== "" && etag === entry.etag) {
-      logger.debug(`${fetchUrl} already present`);
-      releaseLock();
-      return filePath;
-    }
+    // if (entry && etag !== "" && etag === entry.etag) {
```
dbook13, line 44: Is this intentional? Can you add a comment if so?

dbook13 (approved):

Que3216 (approved):

packages/downloading-helpers/src/scripts/replay-assets.ts
alexivanov: that that not intended, dropped in https://github.com/alwaysmeticulous/meticulous-sdk/pull/473/commits/abc34d9e2253a6c781e3bc84b2873f366dbe08a2

PR #474, opened 2023-08-22
drop execution / screenshotting options defaulting from SDK

alexivanov:
This is now BE-driven.

Blocked on the BE part of this going live.


dbook13 (approved):

packages/client/src/api/apply-default-execution-options-from-project.ts
```diff
@@ -1,53 +0,0 @@
...
-
-  if (
-    executionOptions.networkStubbingMode != null ||
-    project.settings.networkStubbingMode == null
-  ) {
-    return executionOptions;
-  }
-
-  if (project.settings.networkStubbingMode.type === "stub-non-ssr-requests") {
-    logger.info("");
```
Que3216, line 31: Note: we may want to keep this logger information somewhere, since I think it is useful feedback when using the CLI.

alexivanov: This is moved to the bundles

PR #478, opened 2023-08-31
Throw instead of process.exit(1)

Que3216:
process.exit bypasses any try/catches if using this as a utility methods (was originally OK-ish since was only used by CLI originally)


Que3216:
Merging to unblock session selection

PR #484, opened 2023-09-14
Deprecate requestsBeforeNetworkRecordingStarted

alexivanov:
This field isn't set within the session data for newly recorded sessions.

### DIFF
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
@@ -25,6 +25,9 @@ export interface SessionData {
   hostname: string;
   abandoned: boolean;
 
+  /**
+   * @deprecated This isn't set for new sessions.
+   */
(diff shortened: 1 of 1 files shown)


packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -25,6 +25,9 @@ export interface SessionData {
   hostname: string;
   abandoned: boolean;
 
+  /**
+   * @deprecated This isn't set for new sessions.
+   */
```
Que3216, line 30: I thought it was set in insertable-snippet.ts#L393?

alexivanov: https://github.com/alwaysmeticulous/meticulous/pull/1979 drops this from the session data.

Que3216: Ah got it! This looks good then

PR #493, opened 2023-10-02
Add Vercel protection bypass token to types

Que3216:


packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
```diff
@@ -178,6 +178,20 @@ export interface ReplayExecutionOptions {
    * This flag is useful to reduce noise when debugging.
    */
   essentialFeaturesOnly: boolean;
+
+  vercel?: {
+    deploymentProtectionBypassToken: string | null;
+  };
```
Que3216, line 184: Decided not to do a generic 'additionalHeaders' option, since we'll only want to send this header to origins we know are safe to send it to.

PR #500, opened 2023-10-11
add class name hashes difference screenshot diff result type

alexivanov:


Que3216 (approved):

packages/api/src/sdk-bundle-api/bundle-to-sdk/screenshot-diff-result.ts
```diff
@@ -81,6 +80,18 @@ export interface ScreenshotDiffResultCompared {
...
+export interface ScreenshotDiffResultNoDifference
+  extends ScreenshotDiffResultCompared {
+  outcome: "no-diff";
+}
+
+export interface ScreenshotDiffResultDifference
+  extends ScreenshotDiffResultCompared {
+  outcome: "diff";
+
+  hashOfChangedSectionsClassNames?: string;
```
Que3216, line 92: [nit / not important / can skip] Could always add a comment here to explain what this is / how it's computed

PR #508, opened 2023-10-23
Don't record initial docs page session

alexivanov:


packages/record/src/record/record.utils.ts
```diff
@@ -33,11 +33,14 @@ export async function bootstrapPage({
   await page.evaluateOnNewDocument(earlyNetworkRecorderSnippetFile);
 
   page.on("framenavigated", async (frame) => {
-    if (page.url() === INITIAL_METICULOUS_DOCS_URL) {
```
dbook13, line 36: Why did the old way of doing it not prevent recording the initial docs page?

dbook13 (approved):

packages/record/src/record/record.utils.ts
alexivanov: We do record sessions on the app for our internal testing so the snippet got injected during the initial navigation too, which isn't desirable here.

PR #514, opened 2023-11-02
Store head and base replay id for flakes

Que3216:
Without the head replay id it's impossible to find / load the correct screenshots for the flake retry


Que3216:
Merging to unblock

PR #522, opened 2023-11-06
New type for customRequestHeaders

dbook13:
Create a stronger type for customRequestHeaders which allows headers to be injected in different ways with different restrictions. The flexibility introduced here will support user-defined headers to be passed through to certain requests (unblocks AssetNote) and will allow us to consolidate the vercel bypass token header injection logic with the rest of our header injection logic.

For a WIP example of what these types will look like in use, see https://github.com/alwaysmeticulous/meticulous/pull/2165/files#diff-8ebdbdeb9d048873a79c40d6b5ae9eaa0f4837bc751a552176288a19b6c59769R98


packages/api/src/sdk-bundle-api/sdk-to-bundle/header-injection.ts
```diff
@@ -0,0 +1,39 @@
...
+  requestTargets: AllRequests | AppUrlRequestsOnly | CustomRequests;
+}
+  
+export interface StaticHeaderValue {
+  type: "static";
+  value: string;
+}
+
+// Note: only for internal use initially, but we might expose this type to users in the future.
+export interface DynamicHeaderValue {
```
dbook13, line 15: Needed for `Meticulous-Simulated-Date` currently.

packages/api/src/sdk-bundle-api/sdk-to-bundle/header-injection.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface AllRequests {
+  type: "all";
+}
+
+export interface AppUrlRequestsOnly {
+  type: "app-url-only";
+}
+
+export interface CustomRequests {
```
dbook13, line 28: Not necessary for AssetNote but provides a lot of additional flexibility.

packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
```diff
@@ -184,7 +185,7 @@ export interface ReplayExecutionOptions {
   /**
    * If populated, each header will be injected into all requests when fetching resources during a replay.
    */
-  customRequestHeaders?: Record<string, string>;
+  customRequestHeaders?: InjectableRequestHeader[];
```
dbook13, line 188: The readme mentions that changing a type is unsafe but since this has never been used I think it's fine?

alexivanov (approved):

PR #531, opened 2023-11-17
Support capturing HTTP only cookies

Que3216:


packages/record/src/record/utils/provide-cookie-access.ts
```diff
@@ -0,0 +1,43 @@
+import { Cookie } from "@alwaysmeticulous/api";
+import { Page, Protocol } from "puppeteer";
+
+type ModifiedWindow = {
+  __meticulous?: {
+    getAllCookies?: () => Promise<Cookie[]>;
```
Que3216, line 6: This is called by the recorder script if present

packages/record/src/record/utils/provide-cookie-access.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+};
+
+export const provideCookieAccess = async (page: Page) => {
+  const client = await page.target().createCDPSession();
+  const cookies = (await client.send("Network.getAllCookies")).cookies.map(
```
alexivanov, line 12: Does this differ from `Page.cookies()`?

packages/cli/src/commands/record/record.command.ts
```diff
@@ -188,5 +191,10 @@ export const recordCommand = buildCommand("record")
       boolean: true,
       description: "Enable verbose logging",
     },
+    captureHttpOnlyCookies: {
+      boolean: true,
+      default: true,
```
alexivanov, line 196: I wonder if the implicit nature of this divergence between the CLI and the snippet might lead to confusion down the line for cases where simulations work on session recorded via the CLI during onboarding / testing but not for passively recorded ones.

alexivanov (approved):

packages/record/src/record/utils/provide-cookie-access.ts
Que3216: I beleive that one doesn't include http only cookies https://stackoverflow.com/questions/49389775/puppeteers-page-cookies-not-retrieving-all-cookies-shown-in-the-chrome-dev-to

packages/cli/src/commands/record/record.command.ts
Que3216: Yeah I was worried about that... but at least if it works with CLI sessions but not with normal ones we'd be able to guess why. Chrome extensions [can access httpOnly cookies](https://developer.chrome.com/docs/extensions/reference/cookies/), if we have robot auto-genned tests they will be able to do so too, so my thinking here was that maybe we default this to true, instead of defaulting to the lowest common denominator?

PR #532, opened 2023-11-17
Only download metadata files if they exist

Que3216:


Que3216:
Merging so that can release & bump in one go / to fix bug

PR #540, opened 2023-12-07
Add session storage to types

Que3216:


packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -11,9 +11,19 @@ export interface SessionData {
     pollyHAR?: { [recordingId: `Meticulous_${string}`]: { log: HarLog } };
   };
 
+  /**
+   * Note: the name 'randomEvents' is a misnomer: it should be named 'storage'.
+   */
   randomEvents: {
```
Que3216, line 17: At some point if we do a session data migration we can rename this, but I don't think it's worth it right now

dbook13 (approved):
Nice

PR #551, opened 2024-01-17
Fetch sourcemaps when running against local snippet server

edoardopirovano:
This will facilitate debugging by allowing us to attach a debugger and see the actual sources when debugging changes to the snippets.

### DIFF
--- a/packages/downloading-helpers/src/scripts/replay-assets.ts
+++ b/packages/downloading-helpers/src/scripts/replay-assets.ts
@@ -26,7 +26,10 @@ const ASSET_METADATA_FILE_NAME = "assets.json";
 export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
   const fetchUrl = new URL(path, getSnippetsBaseUrl()).href;
-  const assetFileName = `${basename(new URL(fetchUrl).pathname, ".js")}.cjs`;
+  let assetFileName = basename(new URL(fetchUrl).pathname);
+  if (assetFileName.endsWith(".js")) {
+    assetFileName = `${assetFileName.slice(0, -3)}.cjs`;
+  }
   const client = axios.create();
   axiosRetry(client, { retries: 3 });
 
@@ -44,10 +47,14 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
     if (entry && etag !== "" && etag === entry.etag) {
       logger.debug(`${fetchUrl} already present`);
       releaseLock();
+      await fetchSourceMapIfDebugging(fetchUrl);
       return filePath;
     }
 
-    const contents = (await client.get(fetchUrl)).data;
+    let contents = (await client.get(fetchUrl)).data;
+    if (path.endsWith(".map")) {
+      contents = JSON.stringify(contents);
+    }
     await writeFile(filePath, contents);
     if (entry) {
       logger.debug(`${fetchUrl} updated`);
@@ -58,6 +65,7 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
     }
     await saveAssetMetadata(assetMetadata);
     releaseLock();
+    await fetchSourceMapIfDebugging(path);
     return filePath;
   } catch (err) {
     releaseLock();
@@ -65,6 +73,15 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
     throw err;
   }
 };
+
+async function fetchSourceMapIfDebugging(path: string) {
+  if (
+    process.env.METICULOUS_SNIPPETS_BASE_URL?.includes("localhost") &&
+    path.endsWith(".js")
+  ) {
+    await fetchAsset(`${path}.map`);
+  }
+}
 
 const getOrCreateAssetsDir: () => Promise<string> = async () => {
   const assetsDir = join(getMeticulousLocalDataDir(), ASSETS_FOLDER_NAME);
--- a/packages/downloading-helpers/src/scripts/replay-assets.ts
+++ b/packages/downloading-helpers/src/scripts/replay-assets.ts
@@ -23,13 +23,30 @@ interface AssetMetadata {
 const ASSETS_FOLDER_NAME = "assets";
 const ASSET_METADATA_FILE_NAME = "assets.json";
 
-export const fetchAsset: (path: string) => Promise<string> = async (path) => {
-  const logger = log.getLogger(METICULOUS_LOGGER_NAME);
+/**
+ * Downloads the given 'module' JS file as an MJS file (to force correct interpretation when loading the JS file).
+ *
+ * The associated source map will also be downloaded if present to sit alongside the main JS file and if the base snippets URL is a localhost URL.
+ */
+export const fetchAsset = async (path: string): Promise<string> => {
+  const snippetsBaseUrl = getSnippetsBaseUrl();
   const fetchUrl = new URL(path, getSnippetsBaseUrl()).href;
-  let assetFileName = basename(new URL(fetchUrl).pathname);
-  if (assetFileName.endsWith(".js")) {
-    assetFileName = `${assetFileName.slice(0, -3)}.cjs`;
+  const assetFileName = basename(new URL(fetchUrl).pathname);
+  const assetFileNameAsCjsFile = convertJsExtensionToCJS(assetFileName);
+
+  const jsFilePath = await fetchAndCacheFile(fetchUrl, assetFileNameAsCjsFile);
+  if (snippetsBaseUrl.includes("localhost")) {
+    await fetchAndCacheFile(`${fetchUrl}.map`, `${assetFileName}.map`);
   }
+
+  return jsFilePath;
+};
+
+const fetchAndCacheFile = async (
+  urlToDownloadFrom: string,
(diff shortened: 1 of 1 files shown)


packages/downloading-helpers/src/scripts/replay-assets.ts
```diff
@@ -66,6 +74,15 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   }
 };
 
+async function fetchSourceMapIfDebugging(path: string) {
+  if (
+    process.env.METICULOUS_SNIPPETS_BASE_URL?.includes("localhost") &&
```
Que3216, line 79: [nit] Could use `getSnippetsBaseUrl()` if want to keep the magic constant name in one place:

```suggestion
    getSnippetsBaseUrl().includes("localhost") &&
```

packages/downloading-helpers/src/scripts/replay-assets.ts
```diff
@@ -58,6 +65,7 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
     }
     await saveAssetMetadata(assetMetadata);
     releaseLock();
+    await fetchSourceMapIfDebugging(path);
```
Que3216, line 68: [nit] I found the recursion made it a bit slower for me to parse what's going on. For fast readability you could also rename fetchAsset to fetchJsFile, have fetchAsset just download an arbitrary asset form a URL, and then communicate the intent more directly by doing:

```ts
/**
 * Downloads the given 'module' JS file as an MJS file (to force correct interpretation when loading the JS file).
 *
 * The associated source map will also be downloaded if present to sit alongside the main JS file and if the base snippets URL is a localhost URL.  
 */
export const fetchJsFile = (path: string): Promise<string> => {
     const snippetsBaseUrl = getSnippetsBaseUrl();
     const fetchUrl = new URL(path, getSnippetsBaseUrl()).href;
     const assetFileName = basename(new URL(fetchUrl).pathname);
     
     const jsFilePath = await fetchAsset(fetchUrl, convertJsExtensionToCjs(assetFileName));
     if (snippetsBaseUrl.includes(localhost)) {
           await fetchAsset(`${fetchUrl}.map`, `${assetFileName}.map`);
     }
     
     return jsFilePath;
}

const fetchAsset = (urlToDownloadFrom: string, fileNameToDownloadAs: string): Promise<string> => {
  // ... original version of fetch asset but without .js specific code: just generic utility for downloading & caching assets ...
}
```

fetchAsset would then be private to the file / not exported

This avoids the reader having to follow the recursion in their mind, convince themselves that it only goes once deep etc. (as soon as they see the recursion they'll probably have questions around at what point does it terminate). The code above communicates the intent and pattern more directly: download the .js file, and then maybe download the .map file.

It also has the slight advantage of colocating the stuff around snippetsBaseUrl / code that changes together lives together. And the seperation is clean: unless they are interested in the caching, most people interested in the code probably won't need to read fetchAsset at all, just fetchJsFile to see what it does.

Side note: fetchAsset is used in the main Meticulous repo as well, but that's OK: we can just switch to the new method name next time we bump the main repo.

Que3216 (approved):
Added a couple of nit comments, but looks great. It'll be nice to get source maps support.

packages/downloading-helpers/src/scripts/replay-assets.ts
```diff
@@ -44,10 +47,14 @@ export const fetchAsset: (path: string) => Promise<string> = async (path) => {
...
       logger.debug(`${fetchUrl} already present`);
       releaseLock();
+      await fetchSourceMapIfDebugging(fetchUrl);
       return filePath;
     }
 
-    const contents = (await client.get(fetchUrl)).data;
+    let contents = (await client.get(fetchUrl)).data;
+    if (path.endsWith(".map")) {
+      contents = JSON.stringify(contents);
```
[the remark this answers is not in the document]
edoardopirovano: The library we use to make the HTTP request is clever enough to realise the body of the response is a JSON object, and it turns it into a JS object rather than a string, which causes errors further down when we try to write out its contents because that code is expecting a string.

packages/downloading-helpers/src/scripts/replay-assets.ts
Que3216: Also, wonder if we should make it not fail if the .map file doesn't exist?

packages/downloading-helpers/src/scripts/replay-assets.ts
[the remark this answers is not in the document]
Que3216: Ah, that's a bit too clever! I'd either add a comment to explain, or see if Axois has another proprety for raw string data (maybe `.data` is not what we want?)

`if (typeof path !== "string")` could also be more direct: that's what we actually care about.

But I expect Axois `.data` is not the field we want.

packages/downloading-helpers/src/scripts/replay-assets.ts
```diff
@@ -26,7 +26,10 @@ const ASSET_METADATA_FILE_NAME = "assets.json";
 export const fetchAsset: (path: string) => Promise<string> = async (path) => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
   const fetchUrl = new URL(path, getSnippetsBaseUrl()).href;
-  const assetFileName = `${basename(new URL(fetchUrl).pathname, ".js")}.cjs`;
+  let assetFileName = basename(new URL(fetchUrl).pathname);
+  if (assetFileName.endsWith(".js")) {
+    assetFileName = `${assetFileName.slice(0, -3)}.cjs`;
+  }
```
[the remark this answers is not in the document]
edoardopirovano: Seems sensible, done!

packages/downloading-helpers/src/scripts/replay-assets.ts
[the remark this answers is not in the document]
edoardopirovano: Unfortunately there isn't a better field! But I agree that checking the type (I believe you mean of `contents` rather than `path`) communicates the intent better. I've also added a comment to explain why we need the type check.

packages/downloading-helpers/src/scripts/replay-assets.ts
edoardopirovano: Aha, yeah that's better. Done.

packages/downloading-helpers/src/scripts/replay-assets.ts
edoardopirovano: This is indeed much nicer and avoids the recursion, done. I went for `fetchAsset` and `fetchAndCacheFile` so that we avoid the breaking change in the name. 

> Also, wonder if we should make it not fail if the .map file doesn't exist?

I'd prefer to not. Obviously it means a slight annoyance that we will fail if the snippet server is an older version that isn't serving sourcemaps. But I think keeping things simple and making sure stuff breaks loudly if in the future if we're missing source maps (e.g. we change our build and it stops producing them in the right place) is better.

packages/downloading-helpers/src/scripts/replay-assets.ts
```diff
@@ -23,45 +23,74 @@ interface AssetMetadata {
...
+export const fetchAsset = async (path: string): Promise<string> => {
+  const snippetsBaseUrl = getSnippetsBaseUrl();
   const fetchUrl = new URL(path, getSnippetsBaseUrl()).href;
-  const assetFileName = `${basename(new URL(fetchUrl).pathname, ".js")}.cjs`;
+  const assetFileName = basename(new URL(fetchUrl).pathname);
+  const assetFileNameAsCjsFile = convertJsExtensionToCJS(assetFileName);
+
+  const jsFilePath = await fetchAndCacheFile(fetchUrl, assetFileNameAsCjsFile);
+  if (snippetsBaseUrl.includes("localhost")) {
+    await fetchAndCacheFile(`${fetchUrl}.map`, `${assetFileName}.map`);
```
Que3216, line 39: I think for this one you want?

```suggestion
    await fetchAndCacheFile(`${fetchUrl}.map`, `${assetFileNameAsCjsFile}.map`);
```

Que3216: Since the browser will load 'meticulous.bundle.cjs' and look for 'meticulous.bundle.cjs.map'

packages/downloading-helpers/src/scripts/replay-assets.ts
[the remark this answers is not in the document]
Que3216: You can also avoid the let here too if you like: `await writeFile(filePath, typeof contents !== "string" ? JSON.stringify(contents) : contents);`, but also fine as is.

Que3216 (approved):

packages/downloading-helpers/src/scripts/replay-assets.ts
edoardopirovano: It's actually not! I'm not exactly sure _why_ but both the Chrome debugger and the VS Code one seem to look in `.js.map` even if the file is a `.cjs` one.

packages/downloading-helpers/src/scripts/replay-assets.ts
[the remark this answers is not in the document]
edoardopirovano: Let's leave this one, I think having a separate line here improves readability a bit and gives us somewhere to put the comment.

PR #554, opened 2024-01-18
Add early network recorder loader method

Que3216:
See jsdoc comment for when this is useful. I don't know how common this will be / whether worth supporting, but feels like it will continue to come up?

It will also allow us to have a more performant production recording setup if we can get the early network recorder snippet to be smaller (currently it's 0.5mb, while full snippet is 1.2mb).


packages/recorder-loader/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const SNIPPET_BASE_URL = "https://snippet.meticulous.ai";
```
alexivanov, line 1: tiny nit - I believe we use plural elsewhere in the codebase.
```suggestion
export const SNIPPETS_BASE_URL = "https://snippet.meticulous.ai";
```

packages/recorder-loader/src/install-meticulous-intercepts.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (stopFunction && !stoppedRecording) {
+      await stopFunction();
+      stoppedRecording = true;
+    }
+  };
+  const startRecordingSession = tryLoadAndStartRecorder;
+  const interceptor = { startRecordingSession, stopIntercepting };
+
+  const promise = new Promise<Interceptor>((resolve, reject) => {
+    const timeout = setTimeout(() => {
```
alexivanov, line 67: maybe we should only set timeout if `options.maxMsToBlockFor) > 0` similar to `tryLoadAndStartRecorder`, to enable disabling this behaviour (which also isn't well documented)

alexivanov (approved):

PR #555, opened 2024-01-19
Download diffs too when downloading a replay

edoardopirovano:
This will marginally improve my debugging experience by allowing me to download diff screenshots using the CLI. It also means we handle a return value containing the diffs, which will allow us to start actually returning this from the backend.

### DIFF
--- a/packages/downloading-helpers/src/scripts/replays.ts
+++ b/packages/downloading-helpers/src/scripts/replays.ts
@@ -124,7 +124,7 @@ const downloadReplayV3Files = async (
     );
   }
 
-  const { screenshots, snapshottedAssets, ...rest } = downloadUrls;
+  const { screenshots, diffs, snapshottedAssets, ...rest } = downloadUrls;
 
   await mkdir(join(replayDir, "screenshots"), { recursive: true });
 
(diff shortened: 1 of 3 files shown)


packages/downloading-helpers/src/scripts/replays.ts
```diff
@@ -159,6 +159,32 @@ const downloadReplayV3Files = async (
...
   });
 
+  const diffsFolder = join(replayDir, "diffs");
+  await Promise.all(
+    Object.keys(diffs ?? {}).map((baseReplayId) =>
+      mkdir(join(diffsFolder, baseReplayId), { recursive: true })
+    )
+  );
+
+  const diffsPromises = Object.values(diffs ?? {}).flatMap((diffsForBase) => {
```
Que3216, line 169: The downloadScope === "screenshots" was initially added for [this line of code](https://github.com/alwaysmeticulous/meticulous/blob/847757ae40e6fb584cff3fd24c1b814ccc4ee284/packages/replay-orchestrator/src/replay/screenshot-diffing/download-and-diff-screenshots.ts#L189) so it'd only download the files needed when diffing screenshots. It looks like it isn't being used yet (didn't finish the swing on that one!), but might as well make this `const diffPromises = downloadScope === "everything" ? ... : []`, like line 188, since 'download-and-diff-screenshots' won't care about diff images.

Que3216 (approved):
Nice!

packages/downloading-helpers/src/scripts/replays.ts
Que3216: (we probably should update that call to pass the download scope, I suspect it'd make a reasonable perf difference)

Que3216: Opened a PR here https://github.com/alwaysmeticulous/meticulous/pull/2431

edoardopirovano: Thanks, done! And approved your PR 👍🏼

PR #557, opened 2024-01-25
Bump puppeteer version

alexivanov:
Looks like 21.9.0 fixes the "aw snap" issues we're seeing when recording.

This also bumps the simulations Chrome version, I'm planning on bumping cloud replay Chrome version / the GH action version after this is merged.

### Links
See https://github.com/puppeteer/puppeteer/issues/11672

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -40,7 +40,7 @@
     "@alwaysmeticulous/sentry": "^2.72.0",
     "@sentry/node": "^7.36.0",
     "loglevel": "^1.8.0",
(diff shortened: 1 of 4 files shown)


Que3216:
Nice! Wonder if we should bump ENVIRONMENT_VERSION when we bump the main repo?

Que3216 (approved):

alexivanov:
>Wonder if we should bump ENVIRONMENT_VERSION when we bump the main repo?

Most likely, yes. Will check out if the internal tests show diffs due to the bump

PR #560, opened 2024-01-26
bypass CSP for record-login by default

alexivanov:


packages/cli/src/commands/record-login/record-login.command.ts
```diff
@@ -81,9 +81,17 @@ export const recordLoginFlowCommandHandler: (
...
+export const recordLoginCommand = buildCommand("record-login")
   .details({
     describe: "Record a login flow session",
   })
-  .options(COMMON_RECORD_OPTIONS)
+  .options({
+    ...COMMON_RECORD_OPTIONS,
+
+    // We explicitly set the default to true for this command.
+    bypassCSP: {
```
dbook13, line 92: Ah cool, I didn't realize we already had a way to do this.

dbook13 (approved):

PR #562, opened 2024-01-30
[record-login] Wait for session data to save before exiting, and add --appUrl parameters

Que3216:
Fixes ENG-833

### DIFF
--- a/packages/record/src/record/record-login-flow.ts
+++ b/packages/record/src/record/record-login-flow.ts
@@ -155,7 +155,7 @@ export const recordLoginFlowSession = async ({
         // Navigate to the final login flow url and flush any pending payloads
         logger.debug("Reloading page to capture application storage...");
         await loginDataPage.goto(finalLoginUrl, {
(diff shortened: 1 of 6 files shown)


packages/record/src/record/record-login-flow.ts
```diff
@@ -148,12 +153,11 @@ export const recordLoginFlowSession = async ({
         });
 
         // Navigate to the final login flow url and flush any pending payloads
+        logger.debug("Reloading page to capture application storage...");
         await loginDataPage.goto(finalLoginUrl, {
-          waitUntil: "domcontentloaded",
-        });
-        await loginDataPage.evaluate(() => {
-          (window as ModifiedWindow).__meticulous?.flushPendingPayloads?.();
+          waitUntil: "load",
```
Que3216, line 158: I thought I hit a case on one website where changing this to "load" caused it to hang indefinitely... but I can't reproduce it, so going to assume it's safe for now and if we notice it hanging we can fix it.

PR #564, opened 2024-02-08
Tone down data dir warnings

Que3216:
The local data dir defaults to ~/.meticulous and this is normally what is desired and relied upon. We only explicitly set the data dir if:

1. The user passes the --dataDir flag via the CLI
2. In execute session task handler and replay task handler (and in these cases it just grabs `getMeticulousLocalDataDir()` in the spawner process and sends that string through to the spawned process)

### DIFF
--- a/packages/common/src/local-data/local-data.ts
+++ b/packages/common/src/local-data/local-data.ts
@@ -7,10 +7,10 @@ let _localDataDir = "";
 export const getMeticulousLocalDataDir: () => string = () => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
   if (!_localDataDir) {
-    logger.warn(
-      "Local data dir has not been set by setMeticulousLocalDataDir()!"
-    );
     setMeticulousLocalDataDir();
+    logger.debug(
+      `Local data dir has not been set explictly, so defaulting to ${_localDataDir}`
+    );
   }
   logger.debug(`Meticulous local data dir: ${_localDataDir}`);
   return _localDataDir;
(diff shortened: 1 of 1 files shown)


packages/common/src/local-data/local-data.ts
```diff
@@ -7,10 +7,10 @@ let _localDataDir = "";
 export const getMeticulousLocalDataDir: () => string = () => {
   const logger = log.getLogger(METICULOUS_LOGGER_NAME);
   if (!_localDataDir) {
-    logger.warn(
-      "Local data dir has not been set by setMeticulousLocalDataDir()!"
-    );
     setMeticulousLocalDataDir();
+    logger.debug(
```
dbook13, line 11: I wonder if we'd be fine with just the log in line 15? I'm fine either way.

dbook13 (approved):

packages/common/src/local-data/local-data.ts
Que3216: I'll make it an else

packages/common/src/local-data/local-data.ts
```diff
@@ -7,10 +7,10 @@ let _localDataDir = "";
...
+      `Local data dir has not been set explictly, so defaulting to ${_localDataDir}`
+    );
   }
   logger.debug(`Meticulous local data dir: ${_localDataDir}`);
```
Que3216, line 15:
```suggestion
  } else {
    logger.debug(`Using local data dir at ${_localDataDir}`);
  }
```

PR #565, opened 2024-02-09
Parallelize by default

Que3216:


packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
```diff
@@ -141,9 +168,9 @@ export const runAllTestsCommand = buildCommand("run-all-tests")
       description: "Outputs a summary page for GitHub actions",
       default: false,
     },
-    parallelize: {
+    noParallelize: {
```
dbook13, line 171: Could we keep the old name and change the default to true? Having the double-negative `!noParallelize` all over the place makes the code a bit more confusing to read through.

dbook13 (approved):

packages/cli/src/commands/run-all-tests/run-all-tests.command.ts
Que3216: I want to optimize for the end-user rather than us as programmers in this case (want to start treating the CLI UX as more first class / same care with which we'd treat the webapp). Though I could certainly alias `const parallelize = !noParallelize` if you think it'd make it easier to read?

I was thinking `--no-parallelize` is easier and maybe even more intuitive for the user than `--parallelize=false`? (I think I've also seen other CLIs use `--no-parallelize`)

PR #569, opened 2024-02-15
Expose flag for setting a session ID for app storage seeing during replay / test run execution

alexivanov:


alexivanov:
>As a side note I wonder if we should have a METICULOUS_POWER_USER env var and we set all the hidden flags to false if that is set to true. Power users can then set that flag in the .bashrc and more easily discover these more advanced features?

That's a good shout, I can do this in a follow up PR.

PR #584, opened 2024-03-04
WebSocket data types

dbook13:
Needed for https://github.com/alwaysmeticulous/meticulous/pull/2598

### DIFF
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
@@ -27,6 +27,11 @@ export interface SessionData {
     };
   };
 
+  /**
+   * Only present on recordings since ~March 2024
+   */
+  webSocketData?: WebSocketConnectionData[];
+
   cookies: Cookie[];
   urlHistory: UrlHistoryEvent[];
   rrwebEvents: unknown[];
@@ -93,4 +98,16 @@ export interface EarlyRequest {
   initiatorType: "fetch" | "xmlhttprequest";
   startTime: number;
   duration: number;
+}
+
+export interface WebSocketConnectionData {
+  id: number;
+  url: string;
+  events: WebSocketConnectionEvent[];
+}
+
+export interface WebSocketConnectionEvent {
+  timestamp: number;
+  type: "created" | "opened" | "message sent" | "message received" | "closed" | "error";
+  data?: string;
 }
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
@@ -112,10 +112,10 @@ export interface WebSocketConnectionEvent {
   /**
    * The time in milliseconds since the start of the session.
    * 
-   * During simulations, we consider the "created" event to have been replayed whenever the browser calls `new WebSocket()`
-   * and all other events are replayed at a time relative to the "created" event's timestamp.
+   * During simulations, we consider the "created" event to have been replayed whenever the browser calls 
+   * `new WebSocket()` and all other events are replayed at a time relative to the "created" event's timestamp.
    * 
-   * E.g. the "opened" event is replayed at ("opened".timestamp - "created".timestamp) milliseconds after the "created" event.
+   * E.g. the "opened" event is replayed at ("opened".timestamp - "created".timestamp) milliseconds after the browser calls `new WebSocket()`.
    */
   timestamp: number;
   type: "created" | "opened" | "message-sent" | "message-received" | "closed" | "error";
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
@@ -100,14 +100,24 @@ export interface EarlyRequest {
   duration: number;
 }
 
+export type SequenceNumber = number;
+
 export interface WebSocketConnectionData {
-  id: number;
+  id: SequenceNumber;
   url: string;
   events: WebSocketConnectionEvent[];
 }
 
 export interface WebSocketConnectionEvent {
+  /**
+   * The time in milliseconds since the start of the session.
+   * 
+   * During simulations, we consider the "created" event to have been replayed whenever the browser calls `new WebSocket()`
+   * and all other events are replayed at a time relative to the "created" event's timestamp.
+   * 
+   * E.g. the "opened" event is replayed at ("opened".timestamp - "created".timestamp) milliseconds after the "created" event.
+   */
   timestamp: number;
-  type: "created" | "opened" | "message sent" | "message received" | "closed" | "error";
+  type: "created" | "opened" | "message-sent" | "message-received" | "closed" | "error";
   data?: string;
 }
--- a/packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
+++ b/packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
@@ -195,6 +195,11 @@ export interface ReplayExecutionOptions {
    * to ensure that we don't compare screenshots that were generated from replays with different project settings.
    */
   projectSettingsHash?: string;
+
+  /**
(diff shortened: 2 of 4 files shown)


packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -94,3 +99,15 @@ export interface EarlyRequest {
...
 }
+
+export interface WebSocketConnectionData {
+  id: number;
+  url: string;
+  events: WebSocketConnectionEvent[];
+}
+
+export interface WebSocketConnectionEvent {
+  timestamp: number;
```
Que3216, line 110: How is this timestamp generated? i.e. is it relative to performance.timeOrigin or absolute? And is it using Date.now() or performance.now() / monotonic clock?

Our current timestamps aren't consistent, but we should probably try and standardize on one approach (see https://github.com/alwaysmeticulous/meticulous-sdk/blob/main/packages/api/src/sdk-bundle-api/bidirectional/replayable-event.ts#L32 for context)

Que3216 (commented):
Added one comment/question but will review later together with the main PR

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
dbook13: I was just using performance.now() in the POC, but haven't opened up a PR for that part of websocket support yet. I  will make sure to standardize with what we use in replayable event when I do open a PR for that.

Que3216: If we match replayable events we should probably match timestampRaw rather than timestamp (timestamp is a bit messed up). The ideal timestamp to use would probably be what you're using now, raw performance.now(), since then when debugging / looking at data it's easy to see how many seconds into the session it is, and then seperately store performance.timeOrigin on the session (while if you add on performance.timeOrigin it's some huge number). However we'd have to trade-off if this is worth the lack of consistency with ReplayableEvent.timestampRaw (and look at what over timestamps we record).

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -94,3 +99,15 @@ export interface EarlyRequest {
   startTime: number;
   duration: number;
 }
+
+export interface WebSocketConnectionData {
+  id: number;
```
Que3216, line 104: This just increments for each subsequent connection? If so we could maybe call it sequenceNumber to be clear (or add a comment). I guess technically since webSocketData is an array it's not needed?

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -94,3 +99,15 @@ export interface EarlyRequest {
...
+  url: string;
+  events: WebSocketConnectionEvent[];
+}
+
+export interface WebSocketConnectionEvent {
+  timestamp: number;
+  type: "created" | "opened" | "message sent" | "message received" | "closed" | "error";
```
Que3216, line 111: Let's standardize on kebab-case to match the other (newer) typings, unless these match some types in the spec

packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
```diff
@@ -195,6 +195,11 @@ export interface ReplayExecutionOptions {
    * to ensure that we don't compare screenshots that were generated from replays with different project settings.
    */
   projectSettingsHash?: string;
+
+  /**
+   * Defalts to false. If true, Meticulous will allow websockets to pass through to the real server.
+   */
+  passThroughWebSockets?: boolean;
```
Que3216, line 202: 👍

Que3216: Conceptually this should probably be part of network stubbing mode, and having it part of that may make things easier if and when we want to expose it in the UI. It'd also pave nicely for future in case users eventually want to disable/enable on a per URL basis, and we want to enable this configurability without breaking API. However we'd have to be careful not to change existing semantics and so be a breaking change. One option would be to [update NetworkStubbingMode](https://github.com/alwaysmeticulous/meticulous-sdk/blob/fb9259223cd79d29aaa8969ea76779c4c01fb605/packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts#L1) to:

```ts
export type NetworkStubbingMode =
  | StubAllRequests
  | StubNonSSRRequests
  | CustomStubbing;

/**
 * The default mode. Stubs all requests, apart from ones for _next/static/ files.
 */
export interface StubAllRequests {
  type: "stub-all-requests";
}

/**
 * Stubs all requests apart from NextJS 13 /app dir requests to render server components, and requests for _next/static/ files.
 *
 * Used for NextJs 13 /app directory & server components.
 */
export interface StubNonSSRRequests {
  type: "stub-non-ssr-requests";
}

export interface CustomStubbing {
  type: "custom-stubbing";

  /*
   * Any request that matches any of these filters will be passed through to the real backend and not stubed.
   *
   * We use an allowlist / opt-in pattern to reduce risk (e.g. disabling stubbing for all requests, and then
   * running a replay against production rather than staging, without realizing that stubbing is disabled)
   */
  requestsToNotStub: RequestFilter[];
}

export interface RequestFilter {
  /**
   * If defined will filter to only match requests to a URL matching the
   * specified regex.
   *
   * Any JS regex that passes https://github.com/tjenkinson/redos-detector is supported.
   */
  urlRegex: string;

  /** Defaults to `{ fetch: true, xhr: true, webSockets: false }` */
  connectionTypes?: ConnectionTypesFilter; // <<< ADDED
}

export interface NoStubbing {
  type: "no-stubbing";
  
  /** Defaults to `{ fetch: true, xhr: true, webSockets: false }` */
  connectionTypes?: ConnectionTypesFilter; // <<< ADDED
}

interface ConnectionTypesFilter { // <<< ADDED
      fetch: boolean;
      xhr: boolean;
      webSockets: boolean;
}
```

However I think we probably wouldn't need to perfectly preserve current semantics, and most people who have completely disabled stubbing probably want to disable web socket stubbing too, and so we could have connectionTypes absent from NoStubbing and it defaults to completely disabling stubbing, and maybe even have the default value for connectionTypes on RequestFilter to be `{ fetch: true, xhr: true, webSockets: true }` which is a more expected default value.

Que3216: My instinct would be to err on the side of making the breaking change and doing:

```ts
export interface RequestFilter {
  ...

  /** Defaults to `{ fetch: true, xhr: true, webSockets: true }` */
  connectionTypes?: ConnectionTypesFilter; // <<< ADDED
}

export interface NoStubbing {
  type: "no-stubbing";
}
```

Que3216 (approved):
A couple of comments but looks great!

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
dbook13: True, the ID is essentially the index in the array here, but for recording ws data I was planning on making each payload a map of id to WebSocketConnectionEvent[]. This structure makes it a bit cleaner to handle multiple ws connections to the same URL as well as payloads were no events were recorded for a connection. We could call this field `index` but I think `id` requires less explanation and makes the data structures a bit more consistent.

I'll update the type to `SequenceNumber`.

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
dbook13: The interesting thing about the ws timestamps is that it's only really the created event where we care about the time relative to the beginning of the session. For every other event, we just care about time relative to the created event.

Maybe because this use of timestamps is different from how we use timestamps for replayable events, it's fine to not be completely consistent?

packages/sdk-bundles-api/src/replay-orchestrator/sdk-to-bundle/execute-replay.ts
dbook13: Extending network stubbing mode and making the breaking change makes sense to me.

PR #585, opened 2024-03-05
Fix concurrency problems in replay download

edoardopirovano:
This code is kind of broken at the moment, although until the waiting for base screenshots work this wasn't being surfaced as we weren't hitting the race conditions in it (or at least, if we were this wasn't being surfaced: it might have subtly caused some missing base screenshots).

The problem here is that we use the `metadata.json` file to decide if the replay has been previously downloaded. But when the scope is set to `screenshots-only` we don't actually download this file. So, if two processes simultaenously try to download just screenshots, they'll both perform the download. The locking makes sure that only one is downloading at a given time, but the second one to arrive will start overwriting the files that the first one might now have gone on to read. This can lead to corrupted data being read.

This should fix the above bug by recording what we've previously downloaded in a new file for this purpose.

### DIFF
--- a/packages/downloading-helpers/src/scripts/replays.ts
+++ b/packages/downloading-helpers/src/scripts/replays.ts
@@ -1,3 +1,4 @@
+import { readFileSync, writeFileSync } from "fs";
 import { mkdir, rm } from "fs/promises";
 import { dirname, join } from "path";
 import {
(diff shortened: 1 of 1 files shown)


Que3216 (approved):
Nice

PR #590, opened 2024-03-11
Create tunnel client

alexivanov:
### Changes

This adds a client for our secure tunnels service.

This client is based on https://github.com/localtunnel/localtunnel with some modifications to make it secure and tailor it for our use case.

This package expose a command for interacting with the client but after the initial testing most of the access will be programatically via the API, e.g. within GitHub Actions or as part of other command handlers. The API will most likely change in following versions as things are still in developmens.

### Code review

Review by commit might be easier, first PR is quite large though but it's just forking the upstream master.


packages/cli/src/commands/start-local-tunnel/start-local-tunnel.command.ts
```diff
@@ -0,0 +1,149 @@
...
+    localHttps: argv.localHttps,
+    localCert: argv.localCert,
+    localKey: argv.localKey,
+    localCa: argv.localCa,
+    allowInvalidCert: argv.allowInvalidCert,
+  }).catch((err) => {
+    throw err;
+  });
+
+  tunnel.on("error", (err) => {
```
dbook13, line 56: Should we call `tunnel.close()` here as well?

packages/cli/src/commands/start-local-tunnel/start-local-tunnel.command.ts
```diff
@@ -0,0 +1,149 @@
...
+    tunnelClosedCallback.resolve(null);
+  });
+
+  logger.info(
+    `Your url is: ${tunnel.url}, user: ${tunnel.basicAuthUser}, password: ${tunnel.basicAuthPassword}`
+  );
+
+  if (argv.printRequests) {
+    tunnel.on("request", (info: IncomingRequestEvent) => {
+      console.log(new Date().toString(), info.method, info.path);
```
dbook13, line 72: logger.info instead?

packages/tunnel-client/src/lib/tunnel.ts
```diff
@@ -0,0 +1,212 @@
...
+  tunnel_passphrase: string;
+  basic_auth_user: string;
+  basic_auth_password: string;
+}
+
+interface CreateTunnelResponseError {
+  error: string;
+}
+
+export class Tunnel extends EventEmitter {
```
dbook13, line 23: Can you set the return types on this class' methods? With all the callbacks being passed around it's a bit confusing.

dbook13 (approved):
Nice!

packages/tunnel-client/src/lib/tunnel.ts
alexivanov: Good shout, https://github.com/alwaysmeticulous/meticulous-sdk/pull/590/commits/32c4fc16363d479e113ec6561f9241fb2da2a73e does that

PR #603, opened 2024-03-25
Add binary data types to WebSocketConnectionMessageEvent

dbook13:
Needed for https://github.com/alwaysmeticulous/meticulous/pull/2660


packages/api/src/sdk-bundle-api/sdk-to-bundle/websocket-data.ts
```diff
@@ -37,10 +37,24 @@ export interface WebSocketConnectionOpenedEvent
...
+export interface EncodedArrayBuffer {
+  binaryType: "arraybuffer";
+  /** Base 64 encoded binary from an array buffer */
+  encodedData: string;
+}
+
+export interface EncodedBlob {
+  binaryType: "blob";
+  /** Base 64 encoded binary from a blob */
+  encodedData: string;
```
Que3216, line 49: Fun fact I learnt today: although base64 uses 33% extra storage, if the network traffic is gzipped the storage penalty almost disappears, leaving only a 2.5-5% penalty (those wasted 2 bits for every 8 get compressed away anyway).

Que3216 (approved):

packages/api/src/sdk-bundle-api/sdk-to-bundle/websocket-data.ts
dbook13: Interesting. Do we gzip session data when sending it to / loading it from S3?

PR #606, opened 2024-03-28
Default or missing port throws an error

fdx-peter:
Fix for when a URL's port matches the default for the protocol, the url.port object is an empty string.
This also happens when a port is not provided in the URL.

See note on:
https://developer.mozilla.org/en-US/docs/Web/API/URL/port

Received the error whilst attempting to run the Github cloud action, with a vanilla url (eg "https://pr.somedomain.com"). The URL parsed to a port of "", where parseInt returned NaN. 

image


Que3216 (approved):
This looks great -- thanks for fixing! As you probably noticed, we just released the local tunnel support, so very new.

We'll need to cut a new release of the SDK and a new release of the GitHub action. We'll let you know when it's all published & ready to use.

alexivanov (approved):

alexivanov:
Thanks @fdx-peter!

PR #619, opened 2024-05-02
Secure tunnel TPC multiplexing

alexivanov:
### Changes

This PR adds a new secure tunnel cluster which works over a single TCP connection and uses a logical multiplexer (https://github.com/davedoesdev/bpmux).

### Rollout

The old 'connection'-based approach (one-to-one mapping of incoming to proxied connections) is kept for backward compatibility during the roll-out period. The type of proxying mode to use is BE-driven -- depending on whether `port` or `multiplexing_remote_port` is set from the API call to create a new tunnel.

### DIFF
--- a/packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts
+++ b/packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts
@@ -139,6 +139,8 @@ export class TunnelMultiplexingCluster extends (EventEmitter as new () => TypedE
     };
 
     remote.on("data", (data: any) => {
+      // parse the first (request) line of the request to determine the method and path
+      // Example: GET /path HTTP/1.1
       const match = data.toString().match(/^(\w+) (\S+)/);
       if (match) {
         this.emit("request", {


packages/tunnel-client/src/lib/tunnel.ts
```diff
@@ -131,25 +144,42 @@ export class Tunnel extends (EventEmitter as new () => TypedEmitter<TunnelEvents
...
+        ...info,
+        multiplexingRemotePort: info.multiplexingRemotePort,
+      });
+    } else if (info.remotePort) {
+      this.logger.debug("using connection agent");
+      tunnelCluster = new TunnelConnectionCluster({
+        ...info,
+        remotePort: info.remotePort,
+        logger: this.logger,
+      });
```
dbook13, line 176: Why do we need to have extra logic before creating the `TunnelMultiplexingCluster` but we can create the `TunnelConnectionCluster` directly?

packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts
```diff
@@ -0,0 +1,156 @@
...
+        stream.pipe(local).pipe(remote);
+
+        local.once("close", (hadError) => {
+          this.logger.debug("local connection closed [%s]", hadError);
+        });
+      });
+    };
+
+    remote.on("data", (data: any) => {
+      const match = data.toString().match(/^(\w+) (\S+)/);
```
dbook13, line 142: Maybe a comment here about what this is trying to match could be useful.

packages/tunnel-client/src/lib/tunnel.ts
alexivanov: There is a slight difference in separation of concern between the two. For `TunnelConnectionCluster` because we open a new TCP connection each time we call `open()` we do auth within the "tunnel cluster" but because `TunnelMultiplexingCluster` works over a single connection I've made it so it depends on an open, authenticated TPC connection over which it creates logical duplexes for communication . I guess an alternative might be to open a connection when TunnelMultiplexingCluster is created. I've opted for the former but do shout if you think an alternative here is cleaner!

packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts
alexivanov: https://github.com/alwaysmeticulous/meticulous-sdk/pull/619/commits/75b7fd3f80d5086061a285ce726467c0e891ef29

PR #621, opened 2024-05-07
Add tunnel client tip to readme

Que3216:


alexivanov (approved):
LGTM, `start-local-tunnel` is currently hidden from the help menu (only shown to super users) but maybe it shouldn't? It was originally created to aid development before we created the other secure tunnel wrappers but might be useful to keep it around.

PR #628, opened 2024-05-20
Set port in host header transformer

alexivanov:
A request to a service running on localhost:8080 would've previously resulted in a request with `Host: localhost` being made. The "default" behaviour is to include the non-standard port as well.


dbook13 (approved):
Nice

PR #634, opened 2024-05-24
Handle tunnel test run retries

Que3216:
If the test run transitions to ExecutionError then instead of exiting immediately we log and message and wait up to 5 minutes for the retry. If it gets retried then we wait for the retry to complete. Note that we will wait a 'wasted' 5 minutes if the test run fails twice in a row, since we only retry once, but we always wait to find a retry -- but that should hopefully be rare / don't think is a huge issue.


dbook13 (approved):
LGTM

PR #651, opened 2024-06-19
Fix project typings

Que3216:
`configurationData` no longer returned by controller endpoints.


packages/api/src/project.types.ts
```diff
@@ -8,7 +7,6 @@ export interface Project {
   organization: Organization;
   name: string;
   recordingToken: string;
-  configurationData: ProjectConfigurationData;
```
Que3216, line 11: I checked and this isn't used anywhere in either repo

PR #655, opened 2024-06-30
Add timeout and retries for downloading files

Que3216:
Previously download file could hang indefinitely if we got the first packet but the stream never finished.

Note that if we want to renable downloading to a temp file and renaming I also wrote the code for that here: https://github.com/alwaysmeticulous/meticulous-sdk/pull/655/files/ea58213c4c463f33fb96a9b8c74b6ed2c5aa5ad5#diff-99fd854d0853315ab1f25b49b081a09cd9f85fbefb902bb8f0d4e5ee26f2f3ca


packages/downloading-helpers/src/file-downloads/download-file.ts
```diff
@@ -1,29 +1,79 @@
...
 
-  return client
-    .request({ method: "GET", url: fileUrl, responseType: "stream" })
-    .then(async (response) => {
-      response.data.pipe(writer);
-      return promisifiedFinished(writer);
+    await new Promise((resolve) => writer.close(resolve));
+
+    if (existsSync(path)) {
+      await rm(path);
```
edoardopirovano, line 60: Could we document that this function isn't safe to call concurrently from different processes (i.e. it should not be called within replays and only at the test run level)? I had tried to make it safe to do this, but reverted that because it broke stuff, and now with this line it's perhaps even less safe.

edoardopirovano (approved):

packages/downloading-helpers/src/file-downloads/download-file.ts
Que3216: Ah yeah that's true. I'll have a think and can document it. My assumption here was that if we've started writing to the file and failed mid way through it's probably already corrupted, so better to just delete it.

edoardopirovano: Alternatively we could revive https://github.com/alwaysmeticulous/meticulous-sdk/pull/646 but fix it. The issue was `fileSync` from `tmp` returned a file path in the system's temporary directory but on cloud replay nodes that was on a different file system to the directory we were downloading into so `renameSync` would throw.

PR #659, opened 2024-07-08
Divergence types

dbook13:
Needed for https://github.com/alwaysmeticulous/meticulous/pull/3125

### DIFF
--- a/packages/sdk-bundles-api/src/replay-orchestrator/bundle-to-sdk/execute-test-run.ts
+++ b/packages/sdk-bundles-api/src/replay-orchestrator/bundle-to-sdk/execute-test-run.ts
@@ -30,7 +30,6 @@ export interface TestRunProgress {
 }
 
 export interface DetailedTestCaseResult extends TestCaseResult {
-  // TODO: rename
-  screenshotDiffResultsByBaseReplayId: Record<string, ScreenshotDiffData>;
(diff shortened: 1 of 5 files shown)


packages/sdk-bundles-api/src/replay-orchestrator/bundle-to-sdk/execute-test-run.ts
```diff
@@ -34,6 +30,6 @@ export interface TestRunProgress {
 }
 
 export interface DetailedTestCaseResult extends TestCaseResult {
```
Que3216, line 32: The SDK only uses ExecuteTestRunResult.status (run-all-tests.command.ts), so if you wanted to reduce future SDK bumps when working on divergence you could move ExecuteTestRunResult, DetailedTestCaseResult etc. out of the SDK and into replay-data-api in the main repo. And then have executeTestRun in the SDK return just `{ status: TestRunStatus }`, and in the main repo wrap executeTestRun to cast the result to it's richer type (ExecuteTestRunResult).

Que3216 (approved):

packages/sdk-bundles-api/src/replay-orchestrator/bundle-to-sdk/execute-test-run.ts
dbook13: Think I'll leave it as is for now, but will make the change if bumping the SDK becomes painful when working on divergence going forward.

PR #660, opened 2024-07-08
Add flake evidence field

Que3216:


packages/api/src/sdk-bundle-api/bundle-to-sdk/screenshot-diff-result.ts
```diff
@@ -166,6 +168,8 @@ export interface ScreenshotDiffResultFlake {
   diffsToHeadScreenshotOnRetries: ScreenshotDiffRetryResult[];
 }
 
+export type FlakeEvidence = "varied-results-on-retry" | "diff-is-known-flake";
```
Que3216, line 171: We'll probably filter out 'diff-is-known-flake' completely from the frontend (convert it to no-diff for userVisibleOutcome)

alexivanov (approved):

PR #665, opened 2024-07-16
Clarify warning for downloadAndExtractFile

dbook13:
Based on the comment for downloadFile, it seems like we need to worry about thread safety for the file where we download the gzip rather than the directory where we unpack the gzip. I hit a bug in https://github.com/alwaysmeticulous/meticulous/pull/3181 because I didn't realize this, so I'm modifying the comment here to make it more clear.

### DIFF
--- a/packages/downloading-helpers/src/file-downloads/download-file.ts
+++ b/packages/downloading-helpers/src/file-downloads/download-file.ts
@@ -89,7 +89,8 @@ export const downloadFile = async (
  *
  * Returns a list of the extracted files.
  *
- * Warning: this function is not thread safe. Do not try downloading a file to a tmpZipFilePath that may already be in use by another process.
+ * Warning: this function is not thread safe. Do not try downloading a file to a `tmpZipFilePath`
+ * that may already be in use by another process.
  */
 export const downloadAndExtractFile: (
   fileUrl: string,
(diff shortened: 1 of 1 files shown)


packages/downloading-helpers/src/file-downloads/download-file.ts
```diff
@@ -89,23 +89,24 @@ export const downloadFile = async (
  *
  * Returns a list of the extracted files.
  *
- * Warning: this function is not thread safe. Do not try downloading a file to an extractPath that may already be in use by another process.
+ * Warning: this function is not thread safe. Do not try downloading a file to a `tmpZipFilePath`
```
dbook13, line 92: Do we also care about having a unique extract path? Haven't seen any thread safety warnings for the `extract` function, so I'm assuming no.

edoardopirovano: The original warning is definitely valid too - that's the issue I'd hit in the past. `extract` will overwrite the file so if someone is using it when it runs they may get corrupt data.

dbook13: Ah ok good to know, I'll update the comment.

Just to clarify, was the issue that multiple gzips containing files with the same name were extracted into the same directory?

edoardopirovano: Right, yeah. Specifically it was with the download of scripts from the snippet server (/S3 bucket) which always go to the same directory. There was a case where a script was downloaded inside replays so multiple replays would trigger the download simulatenously and overwrite each-other's script potentially while other replays were using it.

edoardopirovano (approved):

PR #668, opened 2024-07-24
Add support for non-connection pooling multiplexing secure tunnels

alexivanov:
This creates a new 'multiplexing cluster' for handling tunnels in 'non-pooling' connection mode (the 'cluster' naming here is a bit confusing as this no longer deals with a connection pool but can be renamed in the future).

In this new non-pooling mode new (logical) connections are opened from the secure tunnel server on each incoming request.

This also drops the old "connection"-based class


packages/tunnel-client/src/lib/tunnel-multiplexing-pooling-cluster.ts
```diff
@@ -0,0 +1,161 @@
+import EventEmitter from "events";
```
alexivanov, line 1: This is moved from `packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts` with no changes

packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts
```diff
@@ -141,21 +175,23 @@ export class TunnelMultiplexingCluster extends (EventEmitter as new () => TypedE
...
-          method: match[1],
-          path: match[2],
-        });
-      }
-    });
+    this.remoteMuxClient.on("handshake", (stream) => {
+      stream.on("data", (data: any) => {
+        // parse the first (request) line of the request to determine the method and path
+        // Example: GET /path HTTP/1.1
+        const match = data.toString().match(/^(\w+) (\S+)/);
```
edoardopirovano, line 182: It's probably fine because the format here is likely to be very fixed, but just in case should we make this a bit more robust and handle whitespace at the start of the line and extra spaces between the method and path?

edoardopirovano (approved):

packages/tunnel-client/src/lib/tunnel-multiplexing-cluster.ts
alexivanov: Looks like the spec doesn't allow for trailing whitespaces so I think this should be fine -- this is also used for debug printing of requests when `--printRequests` is provided

PR #670, opened 2024-07-29
Support Indexed DB

dbook13:
Used by https://github.com/alwaysmeticulous/meticulous/pull/3266

### DIFF
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
@@ -136,5 +136,10 @@ export type IDBObjectStoreWithEntries = Omit<
   "serialize" | "deserialize"
 > & {
   createObjectStoreOptions: IDBObjectStoreParameters;
+  /**
+   * Entry values are JSON stringified by default. JSON representable objects, arrays and primitives are supported.
+   * For all other types (e.g. Date, Regex, ArrayBuffer etc.), you can define a custom [de]serialize function on the
+   * object store metadata to dictate how the value should be mapped to/from strings. @see IDBObjectStoreMetadata
+   */
   entries: { key?: SerializedIDBValidKey; value: string }[];
 };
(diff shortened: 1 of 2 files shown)


packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -101,3 +108,33 @@ export interface EarlyRequest {
...
+  deserialize?: (value: string) => any;
+}
+
+/**
+ * Currently we only support string keys, but we may support other types in the future.
+ * Keys are:
+ * - Serialized in recorder/src/storage/storage.ts
+ * - Deserialized in puppeteer-utils/src/browser-context/storage.ts
+ *
+ * @see IDBValidKey for all possible types.
```
dbook13, line 125: `IDBValidKey` is a link to the type definition in editors.

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -101,3 +108,33 @@ export interface EarlyRequest {
...
+  type: "string";
+  serializedKey: string;
+}
+
+export type IDBObjectStoreWithEntries = Omit<
+  IDBObjectStoreMetadata,
+  "serialize" | "deserialize"
+> & {
+  createObjectStoreOptions: IDBObjectStoreParameters;
+  entries: { key?: SerializedIDBValidKey; value: string }[];
```
Que3216, line 139:
```suggestion
  /**
   * Entry values are JSON stringified. JSON representable objects, arrays and primitives are supported,
   * all other types (e.g. Date, Regex, ArrayBuffer etc.) are not yet supported.
   */
  entries: { key?: SerializedIDBValidKey; value: string }[];
```

Que3216 (approved):

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
dbook13: Also mentioned that a customer [de]serialize function can be defined to handle unsupported types.

PR #689, opened 2024-09-10
Handle deployment locking

edoardopirovano:
See also the matching internal PR.

### DIFF
--- a/packages/client/src/api/test-run.api.ts
+++ b/packages/client/src/api/test-run.api.ts
@@ -30,7 +30,7 @@ export interface ExecuteSecureTunnelTestRunOptions {
 }
 
 export interface ExecuteSecureTunnelTestRunResponse {
-  testRun: TestRun;
+  testRun?: TestRun;
   deploymentId: string;
 }
 
@@ -42,7 +42,7 @@ export const executeSecureTunnelTestRun = async ({
   basicAuthPassword,
   environment,
   isLockable,
-}: ExecuteSecureTunnelTestRunOptions): Promise<ExecuteSecureTunnelTestRunResponse | null> => {
+}: ExecuteSecureTunnelTestRunOptions): Promise<ExecuteSecureTunnelTestRunResponse> => {
   const { data } = await client
     .post("test-runs/trigger-secure-tunnel-test-run-v2", {
       headSha,
@@ -67,7 +67,7 @@ export const executeSecureTunnelTestRun = async ({
 
       throw error;
     });
-  return (data as ExecuteSecureTunnelTestRunResponse | null) ?? null;
+  return data as ExecuteSecureTunnelTestRunResponse;
 };
 
 export const getTestRun: (options: {
--- a/packages/client/src/api/test-run.api.ts
+++ b/packages/client/src/api/test-run.api.ts
@@ -26,6 +26,12 @@ export interface ExecuteSecureTunnelTestRunOptions {
   basicAuthUser: string;
   basicAuthPassword: string;
   environment: string;
+  isLockable: boolean;
+}
+
+export interface ExecuteSecureTunnelTestRunResponse {
+  testRun: TestRun;
+  deploymentId: string;
 }
 
 export const executeSecureTunnelTestRun = async ({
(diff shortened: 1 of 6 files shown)


packages/client/src/api/test-run.api.ts
```diff
@@ -59,7 +67,7 @@ export const executeSecureTunnelTestRun = async ({
 
       throw error;
     });
-  return (data as TestRun | null) ?? null;
+  return (data as ExecuteSecureTunnelTestRunResponse | null) ?? null;
```
alexivanov: (see comment on internal repo about return type)

packages/remote-replay-launcher/src/index.ts
```diff
@@ -17,6 +18,7 @@ import { getPort } from "./url.utils";
 
 export { TunnelData } from "./types";
 
+const POLL_LOCK_INTERVAL_MS = 30_000; // 30 seconds
```
alexivanov, line 21: We can make this shorter to avoid keeping the runner on, unless there is a reason for the long polling interval? perhaps:
```suggestion
const POLL_LOCK_INTERVAL_MS = 5_000; // 5 seconds
```

packages/remote-replay-launcher/src/types.ts
```diff
@@ -16,6 +16,7 @@ export interface ExecuteRemoteTestRunOptions {
   onTunnelCreated?: (data: TunnelData) => void;
   onTestRunCreated?: (testRun: TestRun) => void;
   onProgressUpdate?: (testRun: TestRun) => void;
+  onTunnelLocked?: () => void;
```
alexivanov, line 19: nit (can ignore for this): might be useful to also expose a `onTestRunCompleted` optional callback.

packages/remote-replay-launcher/src/index.ts
```diff
@@ -155,6 +151,33 @@ export const executeRemoteTestRun = async ({
...
+    }
+    tunnelUnlocked.resolve();
+    return true;
+  };
+  const alreadyUnlocked = await checkUnlocked();
+  if (!alreadyUnlocked) {
+    tunnelCheckInterval = setInterval(checkUnlocked, POLL_LOCK_INTERVAL_MS);
+    await tunnelUnlocked.promise;
+  }
+
```
alexivanov, line 180: The `keepTunnelOpenPromise` logic and `tunnelCheckInterval` clearing can be moved outside of this to improve readability.
```suggestion

    tunnelUnlocked.resolve();
    return true;
  };
  const alreadyUnlocked = await checkUnlocked();
  if (!alreadyUnlocked) {
    tunnelCheckInterval = setInterval(checkUnlocked, POLL_LOCK_INTERVAL_MS);
    await tunnelUnlocked.promise;
  }

  if (tunnelCheckInterval) {
    clearInterval(tunnelCheckInterval);
  }
  if (keepTunnelOpenPromise) {
    await keepTunnelOpenPromise;
  }

  tunnel.close();

```

packages/remote-replay-launcher/src/index.ts
```diff
@@ -155,6 +151,33 @@ export const executeRemoteTestRun = async ({
 
   const completedTestRun = await testRunCompleted.promise;
 
+  const tunnelUnlocked = defer<void>();
+  let tunnelCheckInterval: NodeJS.Timeout | undefined = undefined;
+  const checkUnlocked = async () => {
+    const isLocked = await getIsLocked({ client, deploymentId });
+    if (isLocked) {
+      onTunnelLocked?.();
```
alexivanov, line 159: Does this need to be periodically called? It might be sufficient for this to be called once and let callers implement their own logic for periodic logging etc.

alexivanov: Also possible to make it a `onTestRunCompleted?: ({testRun: TestRun, keepingTunnelOpen: boolean}) => void;` or similarly-named callback. If keeping as is I think the naming / docstring of this callback can be improved to make it clearer it's called periodically and after the test run has completed.

alexivanov (approved):

packages/remote-replay-launcher/src/types.ts
edoardopirovano: Makes sense to add once we need it, but I'll leave this for now.

packages/remote-replay-launcher/src/index.ts
edoardopirovano: That's much cleaner, thank you.

PR #690, opened 2024-09-11
Add types for custom request transformations

dbook13:
Needed for https://github.com/alwaysmeticulous/meticulous/pull/3499

### DIFF
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
@@ -66,9 +66,15 @@ type TransformableRequestData = Pick<Request, "body" | "method" | "url">;
 
 interface CustomTransformationBase {
   /**
-   * The string to search for in the request component. This can also be a regular expression.
+   * The regex to match against the request component.
    */
-  searchString: string;
+  matchRegex: string;
+  /**
+   * The replacement for any matches with matchRegex.
+   * This can reference groups in the matchRegex, for example:
+   * - matchRegex = `id_(\w*)=[^&]*`, replacement = `id_$1=<redacted>`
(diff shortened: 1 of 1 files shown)


packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
```diff
@@ -53,3 +61,41 @@ export interface ConnectionTypesFilter {
   xhr: boolean;
   webSockets: boolean;
 }
+
+type TransformableRequestData = Pick<Request, "body" | "method" | "url">;
+
+interface CustomTransformationBase {
+  /**
+   * The string to search for in the request component. This can also be a regular expression.
```
Que3216, line 69:
```suggestion
   * The regex to match against the request component.
```

(if you want to search a string I think you'd need to write it as a regex and so escape it)

Que3216: (maybe matchRegex or something would be a better name given this)

packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
```diff
@@ -53,3 +61,41 @@ export interface ConnectionTypesFilter {
...
+   */
+  searchString: string;
+  replacement: string;
```
Que3216, line 72:
```suggestion
  /**
   * The replacement to replace at. This can reference groups in the searchString regex, for example:
   *     - matchRegex = `id_(\w*)=[^&]*`, replacement = `id_$1=<redacted>`
   *     - matchRegex= `id_(?<param_name>\w*)=[^&]*`, replacement = `id_$<param_name>=<redacted>`
   */
  replacement: string;
```

Que3216 (approved):

PR #691, opened 2024-09-12
Disable recording replays and change projectId to recordingToken

Que3216:
We already block the script from loading so that we don't record replays but this can lead to a 'Meticulous recorder failed to initialise.' error being logged, which often acts as a red herring leading developers down the wrong path when debugging replay issues.

### DIFF
--- a/packages/recorder-loader/src/loader.ts
+++ b/packages/recorder-loader/src/loader.ts
@@ -9,6 +9,13 @@ const DEFAULT_MAX_MS_TO_BLOCK_FOR = 2_000;
 export const tryLoadAndStartRecorder = async (
   options: LoaderOptions
 ): Promise<void> => {
+  if (window.Meticulous?.isRunningAsTest) {
+    console.debug(
+      "Running as part of a Meticulous test case, so skipping loading the Meticulous recorder."
+    );
+    return;
+  }
+
   // Try to load the recorder and silence any initialisation error.
   await unsafeLoadAndStartRecorder(options).catch((error) => {
     console.error(error);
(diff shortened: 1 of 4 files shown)


packages/recorder-loader/src/loader.types.ts
```diff
@@ -1,7 +1,12 @@
 import { NetworkResponseSanitizer } from "@alwaysmeticulous/sdk-bundles-api";
 
 export interface LoaderOptions {
-  projectId: string;
+  /**
+   * @deprecated Renamed to recordingToken. Please use the same value but pass it as the recordingToken instead.
+   */
+  projectId?: string;
+  recordingToken: string;
```
alexivanov, line 8: This will be a breaking change but I guess that's okay?

packages/recorder-loader/src/loader.ts
```diff
@@ -45,7 +53,7 @@ const unsafeLoadAndStartRecorder = ({
 
     // Setup configuration
     const typedWindow = window;
-    typedWindow.METICULOUS_RECORDING_TOKEN = projectId;
+    typedWindow.METICULOUS_RECORDING_TOKEN = recordingToken ?? projectId;
```
alexivanov, line 56: Typings make `recordingToken` required so I guess we can just use `recordingToken` here?

alexivanov (commented):
👍  might be useful to also update https://github.com/alwaysmeticulous/meticulous-sdk/blob/main/packages/recorder-loader/README.md

alexivanov (approved):

packages/recorder-loader/src/loader.types.ts
Que3216: Yeah I think it's OK because it should be very clear given the deprecation message and it's only when you bump the NPM dep

packages/recorder-loader/src/loader.ts
Que3216: Oh that's true I guess

Que3216: Actually, since some clients are using JS not TS I'll change this back to how it was before

PR #707, opened 2024-10-14
Add redaction package

Que3216:
See `packages/redaction/README.md` for usage examples

### DIFF
--- a/packages/redaction/src/generic/redact-recursively.ts
+++ b/packages/redaction/src/generic/redact-recursively.ts
@@ -26,6 +26,19 @@ export const redactRecursively = (
      *
      * If a new object is returned (different memory reference) then redactRecursively will
      * replace the existing object with the new object and not recurse inside it.
+     *
+     * In the case where you return the obj passed in unchanged, and so allow redactRecursively
+     * to continue recursing inside the object, you should make sure that your redactObject function
+     * returns quickly, and does NOT take time proportional to the size of the nested object. For example
+     * the below code would be O(n^2) in the size of the object and so extremely unperformant on large objects:
+     *
+     * ```
+     * const redactObject = (obj) => {
+     *   const containsSecret = JSON.stringify(obj).includes("secret"); // O(N) where N is size of obj
+     *   return containsSecret ? { __redacted__: true } : obj;
+     * }
+     * const redacted = redactRecursively(aHugeObject, { redactObject }); // O(N^2) (!)
+     * ```
      */
     redactObject?: (obj: object, jsonPath: string[]) => object;
 
@@ -35,6 +48,11 @@ export const redactRecursively = (
      *
      * If a new array is returned (different memory reference) then redactRecursively will
      * replace the existing array with the new array and not recurse inside it.
+     *
+     * Similar to {@link redactObject} you should make sure that your implement executes
+     * quickly in cases where you return the original object unchanged, since it may be executed
+     * many times over in such a case. See the failure case example in the JSDoc for {@link redactObject}
+     * for more details.
      */
     redactArray?: (arr: unknown[], jsonPath: string[]) => unknown[];
     redactString?: (str: string, jsonPath: string[]) => string;
--- a/packages/redaction/src/generic/redact-recursively.ts
+++ b/packages/redaction/src/generic/redact-recursively.ts
@@ -1,0 +1,105 @@
+/**
+ * Given an object recurses through it's structure, including inside arrays, and applies
+ * the provided redaction functions. For example:
+ *
+ * ```js
+ * // Replace all (non-id) strings with asterixes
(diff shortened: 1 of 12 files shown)


packages/sdk-bundles-api/src/record/middleware.ts
```diff
@@ -79,26 +79,37 @@ export interface RecorderMiddleware {
...
+   *     the correct corresponding saved request stored in the recording / recorded session.
+   *
+   *     For example: if you replace all query string values with "[REDACTED]", and there are multiple distinct
+   *     requests with identical paths but different query string values then Meticulous will not have enough
+   *     information to match them correctly. However if instead you md5 hash all query string values then
+   *     Meticulous would have enough information to match the requests correctly.
+   *
+   *  4. Returning null will cause the request and the corresponding response to be dropped from the payload.
+   *     At replay time if there is no exact match for a request that is transformed to null then the request
+   *     will be failed with 'net::ERR_FAILED'/'Failed to fetch'. rather than automatically trying to find a
```
edoardopirovano, line 111:
```suggestion
   *     will be failed with 'net::ERR_FAILED'/'Failed to fetch' rather than automatically trying to find a
```

edoardopirovano (approved):

packages/redaction/src/generic/redact-recursively.ts
```diff
@@ -0,0 +1,105 @@
...
+          }
+          return redacted;
+        });
+        return hasRedacted ? redactedValueEntries : value;
+      } else {
+        if (redactObject) {
+          const redactedValue = redactObject(value as object, jsonPath);
+          if (redactedValue !== value) {
+            return redactedValue;
+          }
```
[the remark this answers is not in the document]
Que3216: It shouldn't because we don't recursively copy at every level -- we just do a shallow copy, and each of the properties just points to the old value.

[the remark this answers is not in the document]
Que3216: Though you could definitely have a user implement a bad redactObject function e.g.:

```ts
const redactObject = (obj: T) => {
   const containsSecret = recrusivelySearchFor(obj, "secret");
   return containsSecret ? { __redacted__: true } : obj;
}
```

My intent is that users mainly use redactString etc. instead where it's harder to shoot yourself in the foot.

I can add a jsdoc warning.

packages/redaction/src/generic/redact-recursively.ts
```diff
@@ -0,0 +1,105 @@
...
+    redactBigInt,
+    redactBoolean,
+  }: {
+    /**
+     * If the original object is returned then redactRecursively will recurse inside it,
+     * and continue applying the redaction.
+     *
+     * If a new object is returned (different memory reference) then redactRecursively will
+     * replace the existing object with the new object and not recurse inside it.
+     */
```
Que3216, line 29:
```suggestion
    /**
     * If the original object is returned then redactRecursively will recurse inside it,
     * and continue applying the redaction.
     *
     * If a new object is returned (different memory reference) then redactRecursively will
     * replace the existing object with the new object and not recurse inside it.
     *
     * In the case where you return the obj passed in unchanged, and so allow redactRecursively
     * to continue recursing inside the object, you should make sure that your redactObject function
     * returns quickly, and does NOT take time proportional to the size of the nested object. For example
     * the below code would be O(n^2) in the size of the object and so extremely unperformant on large objects:
     *
     * ```
     * const redactObject = (obj) => {
     *   const containsSecret = JSON.stringify(obj).includes("secret"); // O(N) where N is size of obj
     *   return containsSecret ? { __redacted__: true } : obj;
     * }
     * const redacted = redactRecursively(aHugeObject, { redactObject }); // O(N^2) (!)
     * ```
     */
```

packages/redaction/src/generic/redact-recursively.ts
```diff
@@ -0,0 +1,105 @@
...
+    redactObject?: (obj: object, jsonPath: string[]) => object;
+
+    /**
+     * If the original array is returned then redactRecursively will recurse inside it,
+     * and continue applying the redaction.
+     *
+     * If a new array is returned (different memory reference) then redactRecursively will
+     * replace the existing array with the new array and not recurse inside it.
+     */
```
Que3216, line 38:
```suggestion
    /**
     * If the original array is returned then redactRecursively will recurse inside it,
     * and continue applying the redaction.
     *
     * If a new array is returned (different memory reference) then redactRecursively will
     * replace the existing array with the new array and not recurse inside it.
     *
     * Similar to {@link redactObject} you should make sure that your implement executes
     * quickly in cases where you return the original object unchanged, since it may be executed
     * many times over in such a case. See the failure case example in the JSDoc for {@link redactObject}
     * for more details.
     */
```

PR #715, opened 2024-10-18
Tunnel client logic for HTTP2 multiplexing

alexivanov:
### Changes

This creates a new secure tunnel server agent which uses HTTP2 under the hood. It is pretty similar to the already existing no-pool multiplexing one but the multiplexing here is done via the native HTTP2 Node implementation rather than an external library.

### DIFF
--- a/packages/tunnel-client/src/lib/tunnel-http2-cluster.ts
+++ b/packages/tunnel-client/src/lib/tunnel-http2-cluster.ts
@@ -1,0 +1,145 @@
+import EventEmitter from "events";
+import { readFileSync } from "fs";
+import { request } from "http";
+import * as net from "net";
+import { createServer, Http2Server } from "node:http2";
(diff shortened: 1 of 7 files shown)


packages/tunnel-client/src/lib/tunnel-http2-cluster.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      pipeline(req, clientReq, (err) => {
+        if (err) {
+          this.logger.error("Request pipeline error", err);
+        }
+      });
+    });
+
+    this.sockets.forEach((socket) => {
+      this.server.emit("connection", socket);
```
Que3216, line 138: Is this "connection" event used anywhere? I can't work out how these sockets are used...

Que3216 (approved):

PR #728, opened 2024-11-15
Add API for record/replay of custom events

edoardopirovano:
This will unblock one of our customers who wants to record files being dropped onto the browser which isn't something our recorder captures at the moment. The biggest open question is whether we allow an async function as a listener and `await` it.

### DIFF
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
@@ -151,8 +151,11 @@ export type IDBObjectStoreWithEntries = Omit<
 
 export type CustomUserEvent = {
   type: string;
+  /**
+   * The timestamp based on performance.now() in the browser when the event was recorded.
+   */
   timestamp: number;
-  data: any;
+  data: string;
 };
 
 export type CustomData = {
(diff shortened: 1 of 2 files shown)


packages/sdk-bundles-api/src/window-api/public-window-api.ts
```diff
@@ -40,6 +40,11 @@ export interface MeticulousPublicReplayApi {
    * Call this method to retrieve an array of custom data that was recorded during the test run.
    */
   retrieveCustomDataArray(arrayId: string): string[];
+
+  /**
+   * Call this method to add a listener for a custom event that was recorded during the test run.
+   */
+  addCustomEventListener(type: string, callback: (data: any) => void): void;
```
Que3216, line 47:
```suggestion
  addCustomEventListener(type: string, callback: (serializedData: string) => void | Promise<void>): void;
```

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -149,7 +149,14 @@ export type IDBObjectStoreWithEntries = Omit<
   entries: { key?: SerializedIDBValidKey; value: string }[];
 };
 
+export type CustomUserEvent = {
+  type: string;
+  timestamp: number;
+  data: any;
```
Que3216, line 155:
```suggestion
  data: string;
```

packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts
```diff
@@ -149,7 +149,14 @@ export type IDBObjectStoreWithEntries = Omit<
   entries: { key?: SerializedIDBValidKey; value: string }[];
 };
 
+export type CustomUserEvent = {
+  type: string;
+  timestamp: number;
```
Que3216, line 154: This is a performance.now() based timestamp? Maybe add jsdoc to be explicit.

PR #759, opened 2024-12-17
First draft at helper for redacting nested fields

Que3216:


packages/redaction/src/generic/redact-nested-fields/utils/primative-field-names.ts
```diff
@@ -0,0 +1,53 @@
...
+// Convert a single tuple to a key-value pair in an redactor object
+type ToRedactorObject<T> = T extends [infer K, infer V]
+  ? K extends string
+    ? { [P in K]: Redactor<V> }
+    : never
+  : never;
+
+type TuplePairsToRedactorObject<T> = UnionToIntersection<ToRedactorObject<T>>;
+
+// TODO: Add support for type unions (test on Tweet instead of OriginalTweet)
```
Que3216, line 47: TODO: Check it handles correctly when fields have multiple types e.g.

```
interface Test {
  prop: string;
  nested: { prop: string[] }
  nested2: { prop: number; }
  nested3: { prop: { an: "object" } }
}
```

packages/redaction/README.md
```diff
@@ -23,3 +25,151 @@ const middleware = [
...
 ];
 ```
+
+## Utilties for redacting javascript objects
+
+To make it easier to redact large complex javascript types we provide a number of helpers.
+
+### asterixOut
+
+This ensures the redacted text is the same length and has the same line breaks as the original text,
```
edoardopirovano, line 35:
```suggestion
This ensures the redacted text is the same length and has the same whitespace as the original text,
```
I don't think we really guarantee that it has the same linebreaks, since if it's rendered in a non-monowidth font the linebreaks might end up in a different space.

packages/redaction/README.md
```diff
@@ -23,3 +25,151 @@ const middleware = [
...
+
+Redact string intelligently redacts the string by checking for common data formats. This reduces the
+risk of your app error'ing during the replay of Meticulous tests due to it failing to operate on the
+redacted data (e.g. "*******" is not a valid URL):
+
+```ts
+import { redactString } from "@alwaysmeticulous/redaction";
+
+const redacted = redactString("some sensitive text"); // returns "**** ********* ****"
+const redacted2 = redactString("test@example.com"); // returns "____@_______.com"
```
edoardopirovano, line 54: Underscores are not a valid character in domain names so this could break some apps that are checking that the email is valid with a strict regex: https://domainname.shop/faq?id=7&currency=GBP&lang=en

We should probably use a different character?

packages/redaction/README.md
```diff
@@ -23,3 +25,151 @@ const middleware = [
...
+  transformJsonResponse({
+    urlRegExp: /https:\/\/api\.example\.com\/.*/,
+    transform: (data: MyComplexApiType) => {
+      return complexApiTypeRedactor(data);
+    },
+  }),
+];
+```
+
+If you update MyComplexApiType to add a new 'name'string field:
```
edoardopirovano, line 100:
```suggestion
If you update MyComplexApiType to add a new `name` string field:
```

packages/redaction/src/generic/__tests__/redact-string.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  it("redacts URLs correctly", () => {
+    expect(redactString("https://example.com")).toMatchInlineSnapshot(
+      `"https://redacted.com/"`
+    );
+  });
+
+  it("redacts ISO 8601 timestamps correctly", () => {
+    expect(redactString("2024-01-01T00:00:00.000Z")).toMatchInlineSnapshot(
+      `"1970-01-01T00:00:00.000Z"`
```
edoardopirovano, line 18: Should we redact to this to a more recent date like the start of the year or month? I'm imagining something like http://myapp.com/orders?year=2024&month=5 and then we redact everything to be in 1970 so the page shows as empty.

edoardopirovano (approved):

packages/redaction/src/generic/redact-string.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // Check if it matches ISO 8601 format
+  if (str.match(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/)) {
+    return new Date(0).toISOString();
+  }
+
+  // Common number formats (SSNs, phone numbers, currency, etc)
+  if (str.match(/^[$£€¥#]?[\d \-_./]+[%]?$/)) {
+    return str.replace(/\d/g, "0");
+  }
```
edoardopirovano, line 23: Should we also do credit card numbers?

Que3216: Yep I'll add one for this.

Que3216: Actually, this should already be handled by the common number format above. But I'll add a test case.

PR #769, opened 2024-12-27
Do not swallow Axios errors

Que3216:


packages/client/src/errors.ts
```diff
@@ -0,0 +1,92 @@
...
+    message = errorMessage;
+    if (requestAndResponse) {
+      message += `\n\n${requestAndResponse}`;
+    }
+  } else {
+    message = requestAndResponse;
+  }
+
+  console.log("request", error.request);
+  console.log("response", error.response);
```
Que3216, line 34:
```suggestion
```

packages/client/src/api/github-cloud-replay.api.ts
```diff
@@ -31,15 +32,7 @@ export const getGitHubCloudReplayBaseTestRun = async ({
       }
     )
     .catch((error) => {
-      if (isAxiosError(error)) {
-        const errorMessage = error.response?.data?.message;
-
-        if (errorMessage) {
-          throw new Error(errorMessage);
```
Que3216, line 38: @alexivanov I think this may have been why we weren't getting full error information (API call that failed etc.) in reports-diff-action logs

Que3216:
Merging to try out in reports-diff-action

PR #794, opened 2025-02-12
Thread through pullRequestHostingProviderId

edoardopirovano:
No interesting logic here, everything meaningful is in the other two repos.

### DIFF
--- a/packages/remote-replay-launcher/src/types.ts
+++ b/packages/remote-replay-launcher/src/types.ts
@@ -30,6 +30,13 @@ export interface ExecuteRemoteTestRunOptions {
 
   environment: string;
   isLockable: boolean;
+
+  /**
+   * The ID of the pull request in the hosting provider, if one is available.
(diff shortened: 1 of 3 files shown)


packages/remote-replay-launcher/src/types.ts
```diff
@@ -30,6 +30,13 @@ export interface ExecuteRemoteTestRunOptions {
 
   environment: string;
   isLockable: boolean;
+
+  /**
+   * The ID of the pull request in the hosting provider, if one is available.
+   * The backend will try to automatically find one from the commit if this is not provided, but in some
+   * cases (e.g. PRs from forks on GitHub) this may not work, hence we prefer to set this if possible.
```
alexivanov, line 37: I think we should still use the BE as primary source of truth for this
```suggestion
   * cases (e.g. PRs from forks on GitHub) this may not work, hence one can be provided here.
```

alexivanov (approved):
Nice!

PR #796, opened 2025-02-15
Better error message if missing http prefix

Que3216:


packages/remote-replay-launcher/src/url.utils.ts
```diff
@@ -1,4 +1,30 @@
...
+} {
+  let url: URL;
+  try {
+    url = new URL(urlString);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (_error) {
+    throw new Error(`Invalid app URL: '${urlString}'`);
+  }
+
+  if (!["http:", "https:", "ws:", "wss:", "ftp:"].includes(url.protocol)) {
```
Que3216, line 13: Does 'https://' work? Should we throw an error if they try using 'https'?

PR #801, opened 2025-02-19
Detect heavy load while tunnel is running

edoardopirovano:
Should make it more visible to us and users when there is resource contention.

### DIFF
--- a/packages/remote-replay-launcher/src/resource-tracker.ts
+++ b/packages/remote-replay-launcher/src/resource-tracker.ts
@@ -1,0 +1,65 @@
+import os from "os";
+import { TestRun } from "@alwaysmeticulous/client";
+import * as Sentry from "@sentry/node";
+import { Logger } from "loglevel";
+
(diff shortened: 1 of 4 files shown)


packages/remote-replay-launcher/src/resource-tracker.ts
```diff
@@ -0,0 +1,65 @@
+import os from "os";
+import { TestRun } from "@alwaysmeticulous/client";
+import * as Sentry from "@sentry/node";
```
Que3216, line 3: I guess Sentry is already initialized whether called [from the CLI](https://github.com/alwaysmeticulous/meticulous-sdk/blob/main/packages/cli/src/main.ts#L31) or from [reports diff action](https://github.com/alwaysmeticulous/report-diffs-action/blob/main/src/actions/cloud-compute/cloud-compute.ts#L14) so this works.

Que3216 (approved):
Nice!

PR #806, opened 2025-02-20
Bump Sentry

edoardopirovano:
There's a mismatch with the `report-diffs-action` repo which is causing problems so let's bump both to the latest.

### DIFF
--- a/packages/sentry/src/index.ts
+++ b/packages/sentry/src/index.ts
@@ -1,6 +1,5 @@
 import { IS_METICULOUS_SUPER_USER } from "@alwaysmeticulous/common";
 import * as Sentry from "@sentry/node";
-import { addExtensionMethods } from "@sentry/tracing";
 import { Duration } from "luxon";
 
 /**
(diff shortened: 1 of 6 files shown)


packages/sentry/src/index.ts
```diff
@@ -38,8 +34,4 @@ export const initSentry: (
       "SecureTunnelInactiveForAppUrlError",
     ],
   });
-
-  addExtensionMethods();
```
alexivanov, line 42: Do we need to register tracing when initialising Sentry now that this is dropped? https://github.com/getsentry/sentry-javascript/blob/7.47.0/MIGRATION.md/#node

alexivanov (approved):

alexivanov:
Note that this will push the minimum node version to v18 which I think we're already requiring.

packages/sentry/src/index.ts
edoardopirovano: Ah probably. Let me add that in.

edoardopirovano: Ah wait, this was removed in v9 actually and now we don't need anything at all: https://docs.sentry.io/platforms/javascript/guides/hapi/migration/v7-to-v8/#updated-sdk-initialization

PR #807, opened 2025-02-20
Output progress for run-all-tests-in-cloud in notty environments

Que3216:
Some CI runners such as CircleCI will timeout the job early if they see no new output.


packages/cli/src/commands/run-all-tests-in-cloud/run-all-tests-in-cloud.command.ts
```diff
@@ -63,6 +63,8 @@ const handler: (options: Options) => Promise<void> = async ({
       format: `Test Run execution progress |${chalk.cyan(
         "{bar}"
       )}| {percentage}% || {value}/{total} tests executed`,
+      noTTYOutput: true,
+      notTTYSchedule: 60000,
```
Que3216, line 67:
```suggestion
      notTTYSchedule: 30000,
```

PR #825, opened 2025-03-14
Use enum for SessionRelevance

linpengzhang:
ORM is not compatible with typescript union types, so let's use enum to guarantee better type safety in our data layer

This change and particularly ORM compatibility has been tested via `yarn link`

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -21,7 +21,7 @@ export {
   RedactedScreenshotsComparison,
 } from "./sdk-bundle-api/bundle-to-sdk/screenshot-diff-result";
 export {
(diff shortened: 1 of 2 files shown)


packages/api/src/index.ts
```diff
@@ -21,7 +21,7 @@ export {
   RedactedScreenshotsComparison,
 } from "./sdk-bundle-api/bundle-to-sdk/screenshot-diff-result";
 export {
-  SessionRelevance,
+SessionRelevance,
```
edoardopirovano, line 24: Formatting broke.

edoardopirovano (approved):

PR #843, opened 2025-04-17
Fix publish-all-packages job

dbook13:
That job had been failing in the publish workflow since April 2nd b/c a clean build resulted in a modified file and a dirty git state ([commits](https://github.com/alwaysmeticulous/meticulous-sdk/commits/main/)). The diff in this PR is the result of a clean build and should fix the job.


dbook13:
Merging without review b/c it only changes a comment and it lets me publish a change I need elsewhere.

PR #845, opened 2025-04-28
Support identity override transformations

linpengzhang:
This enables redefining the "identity transformation". It is particularly useful if we want to enable not matching with the canonical identity (and start already with a more advanced redaction)


edoardopirovano (approved):
I don't understand the context of why we need this, but the change itself looks okay.

PR #846, opened 2025-04-28
Implement asset uploading

edoardopirovano:
This is one of the steps to allowing customers to upload static assets instead of opening a tunnel.


packages/remote-replay-launcher/src/index.ts
```diff
@@ -49,6 +52,27 @@ export const executeRemoteTestRun = async ({
...
+
+  if (appDirectory) {
+    // If the app URL is not a valid URL, it must be a local directory and we bundle it and upload to S3
+    const { testRun } = await uploadAssetsToS3({
+      folder: appDirectory,
+      client,
+      commitSha,
+      rewrites: rewrites ?? [],
+    });
+    return { testRun };
```
alexivanov, line 68: Note that by exiting early here none of the `onXX` callbacks will be called and the completion of the test run won't be awaited. Given how different the behaviour of the tunnel and non-tunnel variants is, it might be worth refactoring this to extract the common functionality into a separate function and the distinct logic into 2 functions ... or expose 2 functions and make the clients of this package call the correct one.

Either way it will be good to document that some of the options passed here are  incompatible (appDirectory & appUrl) (onTunnelCreated / isLockable / keepTunnelOpenPromise / onTunnelStillLocked / secureTunnelHost and rewrites)

edoardopirovano: Yeah, I guess this made more sense before I addressed Daniel's comment of making appUrl/appDirectory separate parameters because then the caller could be agnostic to what it was passing through. But now it needs to handle the two parameters anyway, it's probably better to make it call two separate functions to make it clear they'll have different behaviour. Let me do that.

packages/remote-replay-launcher/README.md
```diff
@@ -1,5 +1,8 @@
 # Remote Replay Launcher
 
-Exposes a simple API for launching remote(running in Meticulous' cloud) test runs 
+Exposes a simple API for launching remote (running in Meticulous' cloud) test runs
 against an app hosted locally, and exposed via the Meticulous secure tunnels service.
 ([see @alwaysmeticulous/tunnel-client](../tunnel-client))
+
+Also exposes an API for launching a remote run by uploading a bundle to S3 for Meticulous
+to download and use.
```
alexivanov, line 8: nit: not storage specific, and for a bit more clarity on the type of 

```suggestion
Also exposes an API for launching a remote run by uploading static build artefacts to Meticulous
and executing tests against them.
```

packages/remote-replay-launcher/src/upload-assets-to-s3-and-trigger-test-run.ts
```diff
@@ -0,0 +1,155 @@
...
+} from "@alwaysmeticulous/client";
+import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
+import archiver from "archiver";
+import log from "loglevel";
+import {
+  UploadAssetsToS3AndTriggerTestRunOptions,
+  ExecuteRemoteTestRunResult,
+} from "./types";
+
+export const uploadAssetsToS3AndTriggerTestRun = async ({
```
alexivanov, line 26: naming nit: S3 is an purely an implementation detail so no need to expose it via the API name here.

```suggestion
export const uploadAssetsAndTriggerTestRun = async ({
```

packages/cli/src/commands/upload-deployment/upload-deployment.command.ts
```diff
@@ -0,0 +1,72 @@
...
+  } catch (error) {
+    if (isOutOfDateClientError(error)) {
+      throw new OutOfDateCLIError();
+    } else {
+      throw error;
+    }
+  }
+};
+
+export const uploadDeploymentCommand = buildCommand("upload-deployment")
```
alexivanov, line 52: Given this will be the main point for kicking off test runs I think the naming should focus on test run execution rather than implementation detail.

Maybe something along the lines of:
```suggestion
export const uploadArtefactExecuteTestRunCommand = buildCommand("upload-artefact-and-execute-test-run-in-cloud")
```

packages/cli/src/commands/upload-deployment/upload-deployment.command.ts
```diff
@@ -0,0 +1,72 @@
...
+  .details({
+    describe:
+      "Upload a deployment to Meticulous, potentially triggering a test run",
```
alexivanov, line 55:
```suggestion
      "Upload build artefacts to Meticulous, potentially triggering a test run",
```

packages/cli/src/commands/upload-deployment/upload-deployment.command.ts
```diff
@@ -0,0 +1,72 @@
...
+  const commitSha = await getCommitSha(commitSha_);
+
+  if (!commitSha) {
+    logger.error(
+      "No commit sha found, you must be in a git repository or provide one with --commitSha"
+    );
+    process.exit(1);
+  }
+
+  logger.info(`Uploading deployment for commit ${commitSha}`);
```
alexivanov, line 34: naming nit -- 'deployment' is sort of internal naming. Here we're not really deploying anything so 'deployment' is a bit misleading.

```suggestion
  logger.info(`Uploading build artefacts for commit ${commitSha}`);
```

packages/remote-replay-launcher/src/execute-remote-test-run.ts
```diff
@@ -0,0 +1,189 @@
+import {
```
alexivanov, line 1: I assume this is just moved from `packages/remote-replay-launcher/src/index.ts`

packages/remote-replay-launcher/src/index.ts
```diff
@@ -1,189 +1,2 @@
...
-  if (keepTunnelOpenPromise) {
-    await keepTunnelOpenPromise;
-  }
-  tunnel.close();
-
-  return {
-    testRun: completedTestRun,
-  };
-};
+export { executeRemoteTestRun, TunnelData } from "./execute-remote-test-run";
```
alexivanov, line 1:
```suggestion
export { executeRemoteTestRun } from "./execute-remote-test-run";
export { TunnelData } from "./types";
```

packages/remote-replay-launcher/src/types.ts
```diff
@@ -40,5 +42,12 @@ export interface ExecuteRemoteTestRunOptions {
 }
 
 export interface ExecuteRemoteTestRunResult {
-  testRun: TestRun;
+  testRun: TestRun | null;
+}
+
+export interface UploadAssetsToS3AndTriggerTestRunOptions {
```
alexivanov, line 48: (same naming nit)

packages/remote-replay-launcher/src/upload-assets-to-s3-and-trigger-test-run.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    return {
+      testRun: result.testRun ?? null,
+    };
+  } finally {
+    try {
+      unlinkSync(zipPath);
+    } catch {
+      // Ignore errors when deleting the temporary file
```
alexivanov, line 78: let's log just in case

packages/cli/src/commands/upload-deployment/upload-deployment.command.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  try {
+    await uploadAssetsToS3AndTriggerTestRun({
+      apiToken,
+      commitSha,
+      appDirectory,
+      rewrites: JSON.parse(rewrites ?? "[]"),
```
alexivanov, line 41: Where do we validate the rewrites json (the shape). If on the server, do we surface the error to the user here? The expected 'shape' of this argument is quite complex so I'd expect users to hit this

alexivanov (approved):
Minor comments, mostly around naming.

packages/cli/src/commands/upload-deployment/upload-deployment.command.ts
edoardopirovano: Ah this is a great point. There should be validation server-side, but let's also validate here so we can fail early and with a good error.

PR #847, opened 2025-04-28
Add where clause in custom transformations

linpengzhang:
This is useful if we want to redact somebody if something else matches certain conditions.
For example, redacting the request body if the url matches X

### DIFF
--- a/packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
+++ b/packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
@@ -85,7 +85,10 @@ interface CustomTransformationBase {
    * - matchRegex= `id_(?<param_name>\w*)=[^&]*`, replacement = `id_$<param_name>=<redacted>`
    */
   replacement: string;
-  requestComponent: keyof TransformableRequestData;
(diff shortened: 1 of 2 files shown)


packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
```diff
@@ -85,7 +85,10 @@ interface CustomTransformationBase {
    * - matchRegex= `id_(?<param_name>\w*)=[^&]*`, replacement = `id_$<param_name>=<redacted>`
    */
   replacement: string;
-  requestComponent: keyof TransformableRequestData;
```
edoardopirovano, line 88: Do we not still need this to know which component the transformation gets applied to in addition to there `where` clause deciding which requests we will apply it to?

edoardopirovano (approved):

packages/api/src/sdk-bundle-api/sdk-to-bundle/network-stubbing.ts
linpengzhang: yes, good catch; got deleted by mistake

PR #857, opened 2025-05-08
Update custom events typing to reflect older sessions

dbook13:
Matches the pattern for other fields we've added to existing session data objects ([e.g.](https://github.com/alwaysmeticulous/meticulous-sdk/blob/main/packages/api/src/sdk-bundle-api/sdk-to-bundle/session-data.ts#L59)), and this typing would have prevented the issue [here](https://github.com/alwaysmeticulous/meticulous/pull/5049). Makes sure we won't hit a similar issue again although most sessions that would have fallen into this category are probably deselected by now.


linpengzhang (approved):
nice!

PR #858, opened 2025-05-12
Create SessionContext type

edoardopirovano:
Also moves the UserId and UserEmail out of being custom data fields and into here so they can fit in with the rest of the new context API.


linpengzhang (approved):
LGTM

PR #864, opened 2025-05-14
Wait for base before uploading assets

linpengzhang:
This PR enables potentially waiting for base before executing a static assets deployment. If the deployment is not a pull request, we'll not wait (the logics is in the internal end-point)

### DIFF
--- a/packages/remote-replay-launcher/src/upload-assets-and-trigger-test-run.ts
+++ b/packages/remote-replay-launcher/src/upload-assets-and-trigger-test-run.ts
@@ -9,7 +9,9 @@ import {
   getApiToken,
   requestAssetUpload,
   createClient,
+  TestRun,
 } from "@alwaysmeticulous/client";
+import { triggerRunOnDeployment } from "@alwaysmeticulous/client/dist/api/project-deployments.api";
 import { METICULOUS_LOGGER_NAME } from "@alwaysmeticulous/common";
 import archiver from "archiver";
 import log from "loglevel";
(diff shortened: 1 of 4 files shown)


packages/remote-replay-launcher/src/upload-assets-and-trigger-test-run.ts
```diff
@@ -18,6 +20,50 @@ import {
...
+        } seconds waiting for test run`
+      );
+    }
+    await new Promise((resolve) =>
+      setTimeout(resolve, POLL_FOR_BASE_TEST_RUN_INTERVAL_MS)
+    );
+    result = await triggerRunOnDeployment(completeAssetUploadArgs);
+    testRun = result?.testRun;
+    baseNotFound = result?.baseNotFound;
+  }
```
edoardopirovano, line 50: We should have some logging (maybe not every time we go around the loop, but at least when we start waiting and once every ~30s so it doesn't look like things are stuck. Otherwise users might cancel the workflow.

packages/remote-replay-launcher/src/upload-assets-and-trigger-test-run.ts
```diff
@@ -18,6 +20,50 @@ import {
...
+  let result = await completeAssetUpload(completeAssetUploadArgs);
+  let testRun = result?.testRun;
+  let baseNotFound = result?.baseNotFound;
+  while (!testRun && baseNotFound) {
+    if (Date.now() - startTime > POLL_FOR_BASE_TEST_RUN_MAX_TIMEOUT_MS) {
+      throw new Error(
+        `Timed out after ${
+          POLL_FOR_BASE_TEST_RUN_MAX_TIMEOUT_MS / 1000
+        } seconds waiting for test run`
+      );
```
edoardopirovano, line 42: Did you mean to `throw` here? I think you just want to log something and `break` out of the loop so you can reach `// Trigger a test run without waiting for base`.

edoardopirovano (approved):

packages/remote-replay-launcher/src/upload-assets-and-trigger-test-run.ts
linpengzhang: good catch!

PR #866, opened 2025-05-21
Warn if no index.html can be found in static assets

linpengzhang:
Static web applications typically have either an index.html file in their root directory. We'll check for both and warn if neither is found.


packages/remote-replay-launcher/src/upload-assets-and-trigger-test-run.ts
```diff
@@ -97,6 +97,20 @@ export const uploadAssetsAndTriggerTestRun = async ({
     throw new Error(`Directory does not exist: ${resolvedAppDirectory}`);
   }
 
+  // If there are no rewrites, check for an index.html in the root directory
+  if (!rewrites || rewrites.length === 0) {
+    const indexHtmlPath = join(resolvedAppDirectory, "index.html");
+    const indexHtmPath = join(resolvedAppDirectory, "index.htm");
```
edoardopirovano, line 103: `index.htm` won't help if they don't have any rewrites because the default rewrite rule we use rewrites to `index.html`.

edoardopirovano (approved):

PR #899, opened 2025-07-15
Support setting localAddress and localPort in tunnel client

edoardopirovano:
This has been requested by Dropbox who will use it to point us to go through a restricted network interface that has limited access to stuff inside their VPN.

### DIFF
--- a/packages/tunnel-client/src/utils/get-local-address.ts
+++ b/packages/tunnel-client/src/utils/get-local-address.ts
@@ -1,12 +1,10 @@
-const LOCAL_ADDRESS_ENV_VARIABLE = "METICULOUS_LOCAL_ADDRESS";
-const LOCAL_PORT_ENV_VARIABLE = "METICULOUS_LOCAL_PORT";
+const LOCAL_ADDRESS_ENV_VARIABLE =
+  "METICULOUS_TUNNEL_CLIENT_REQUEST_LOCAL_ADDRESS";
 
 export const getLocalAddress = () => {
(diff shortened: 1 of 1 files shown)


packages/tunnel-client/src/utils/get-local-address.ts
```diff
@@ -0,0 +1,12 @@
+const LOCAL_ADDRESS_ENV_VARIABLE = "METICULOUS_LOCAL_ADDRESS";
+const LOCAL_PORT_ENV_VARIABLE = "METICULOUS_LOCAL_PORT";
+
+export const getLocalAddress = () => {
+  const localAddress = process.env[LOCAL_ADDRESS_ENV_VARIABLE];
+  const localPort = process.env[LOCAL_PORT_ENV_VARIABLE];
```
alexivanov: discussed in IRL: I think we can drop the `localPort`

packages/tunnel-client/src/utils/get-local-address.ts
```diff
@@ -0,0 +1,12 @@
+const LOCAL_ADDRESS_ENV_VARIABLE = "METICULOUS_LOCAL_ADDRESS";
```
alexivanov: maybe something like this?
```suggestion
const LOCAL_ADDRESS_ENV_VARIABLE = "METICULOUS_TUNNEL_CLIENT_REQEUST_LOCAL_ADDRESS";
```

alexivanov (approved):

PR #907, opened 2025-07-25
Support proxy when opening socket

edoardopirovano:
For the customer who needs to use the tunnel behind a proxy.

### DIFF
--- a/packages/tunnel-client/src/lib/tunnel.ts
+++ b/packages/tunnel-client/src/lib/tunnel.ts
@@ -333,6 +333,59 @@ export class Tunnel extends (EventEmitter as new () => TypedEmitter<TunnelEvents
     this.emit("close");
   }
 
+  private async createProxyConnection(
+    proxyUrl: string,
+    targetHost: string,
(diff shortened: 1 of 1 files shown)


packages/tunnel-client/src/lib/tunnel.ts
```diff
@@ -333,6 +333,69 @@ export class Tunnel extends (EventEmitter as new () => TypedEmitter<TunnelEvents
...
+    targetHost: string,
+    targetPort: number,
+    useTls: boolean,
+  ): Promise<net.Socket> {
+    const url = new URL(proxyUrl);
+    const proxyHost = url.hostname;
+    const proxyPort =
+      parseInt(url.port) || (url.protocol === "https:" ? 443 : 80);
+
+    const proxySocket = net.connect({
```
alexivanov, line 347: This won't work for HTTPS proxies although the logic above would suggest otherwise. Maybe we handle it correctly here or just throw?

alexivanov (approved):

PR #908, opened 2025-07-28
Use proxy-agent when building Axios client

edoardopirovano:
Should hopefully get us to respect our customer's proxy settings from `HTTPS_PROXY`.


packages/client/src/client.ts
```diff
@@ -28,6 +29,7 @@ export const createClient: (options: ClientOptions) => AxiosInstance = ({
     },
     // 60 seconds default timeout
     timeout: 60_000,
+    httpsAgent: getProxyAgent(),
```
alexivanov, line 32: I guess we can set `httpAgent` as well here

alexivanov (approved):

PR #924, opened 2025-08-29
Implement multi-threaded tunnel client

edoardopirovano:
This changes the tunnel client to run multi-threaded. This seems to improve performance, see some results of a local load test below (left=before, right=after). In particular, the rate of errors at this high load is reduced by over an order of magnitude (from 24.2% to 1.5%).

Screenshot 2025-09-01 at 16 49 13

### DIFF
--- a/packages/tunnel-client/src/lib/tunnel.ts
+++ b/packages/tunnel-client/src/lib/tunnel.ts
@@ -1,19 +1,21 @@
 import { EventEmitter } from "events";
-import net from "net";
-import tls from "tls";
+import cluster, { Worker } from "node:cluster";
+import { cpus } from "node:os";
+import path from "path";
 import { Logger } from "loglevel";
 import fetch from "node-fetch";
 import TypedEmitter from "typed-emitter";
 import { IncomingRequestEvent, LocalTunnelOptions, TunnelInfo } from "../types";
 import { getProxyAgent } from "../utils/get-proxy-agent";
-import { TunnelHTTP2Cluster } from "./tunnel-http2-cluster";
+import { WorkerInitOptions } from "./worker-entrypoint";
 
 const DEFAULT_HOST = "https://tunnels.meticulous.ai";
 
 /**
  * Default number of connections to establish for HTTP2 multiplexing.
+ * Uses the number of CPU cores available.
  */
-const DEFAULT_HTTP2_NUMBER_OF_CONNECTIONS = 2;
+const DEFAULT_HTTP2_NUMBER_OF_CONNECTIONS = cpus().length;
 
 interface CreateTunnelResponse {
   id: string;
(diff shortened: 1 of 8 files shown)


packages/tunnel-client/src/utils/open-socket.ts
```diff
@@ -0,0 +1,171 @@
...
+      remoteHost,
+      multiplexingRemotePort,
+      useTls,
+    );
+  } else if (useTls) {
+    socket = tls.connect({
+      host: remoteHost,
+      port: multiplexingRemotePort,
+      rejectUnauthorized: true,
+      ALPNProtocols: ["meticulous-tunnel"],
```
alexivanov, line 106: Some comments got dropped during refactoring
```suggestion
      // The HTTP2 node implementation requires ALPN set.
      // See https://github.com/nodejs/node/blob/9a9409ff1f45c968173118de4cd37dea784f8ec9/lib/internal/http2/core.js#L3039.
      // The server should respond with the ALPN protocol "meticulous-tunnel".
      ALPNProtocols: ["meticulous-tunnel"],
```

packages/tunnel-client/src/utils/open-socket.ts
```diff
@@ -0,0 +1,171 @@
...
+    });
+  }
+
+  socket.setNoDelay(true);
+
+  socket.on("error", (err: NodeJS.ErrnoException) => {
+    logger.debug("got remote connection error", err.message);
+
+    if (err.code === "ECONNREFUSED") {
+      const connectionError = new Error(
```
alexivanov, line 121: (comment here too)

packages/tunnel-client/src/lib/tunnel.ts
```diff
@@ -266,53 +228,63 @@ export class Tunnel extends (EventEmitter as new () => TypedEmitter<TunnelEvents
...
-          multiplexingRemotePort,
-          tunnelPassphrase,
-          sendAuthOkAck: true,
-        }),
-      ),
-    );
+    cluster.setupPrimary({
+      exec: workerPath,
+      silent: false,
+      execArgv: ["--max-old-space-size=2048"],
```
alexivanov, line 248: Can we make this dynamic? We might overcommit on smaller CI runners which might be bad. Also do we need 2G of heap space? I wonder what the default on CI runners is 🤔

packages/tunnel-client/src/lib/tunnel.ts
```diff
@@ -266,53 +228,63 @@ export class Tunnel extends (EventEmitter as new () => TypedEmitter<TunnelEvents
...
+        ...(localKey ? { localKey } : {}),
+        ...(localCa ? { localCa } : {}),
+        allowInvalidCert,
+        proxyAllUrls,
+        rewriteHostnameToAppUrl,
+        enableDnsCache,
+      };
+
+      const worker = cluster.fork({
+        WORKER_OPTIONS: JSON.stringify(workerOptions),
```
alexivanov, line 271: I think we can probably pass these as messages to improve the security of this given that we're the socket passphrase.

alexivanov (approved):
Nice 🪨

packages/tunnel-client/src/lib/tunnel.ts
edoardopirovano: Let's just drop this actually and let Node use it's default. I don't think it's needed.

PR #936, opened 2025-09-09
Typings for companion assets

edoardopirovano:


packages/api/src/sdk-bundle-api/sdk-to-bundle/companion-assets.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Information about companion assets to serve _together_ with a deployment. Note this
+ * is distinct from a static assets deployment, which is _just_ static assets.
+ */
+export interface CompanionAssetsInfo {
+  /**
+   * The upload ID of the deployment to serve the companion assets from.
+   */
+  uploadId: string;
```
alexivanov, line 9: naming nit: `deploymentUploadId` to match `UploadedAssetsReplayTarget`

alexivanov (approved):
LGTM, although shouldn't this be folded into `ReplayTarget` as a new type, similar to static assets?

PR #945, opened 2025-09-16
feature: use puppeteer-core to avoid installing chrome for cli client…

narobertson42:
…s by default

https://www.loom.com/share/54ae48472d9a4abdac7136ccab49f9fb

### DIFF
--- a/packages/common/src/browser-installer.ts
+++ b/packages/common/src/browser-installer.ts
@@ -1,0 +1,252 @@
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import {
+  install,
+  Browser,
+  detectBrowserPlatform,
+  InstallOptions,
+  InstalledBrowser,
+  resolveBuildId,
+  getInstalledBrowsers,
+} from "@puppeteer/browsers";
+import chalk from "chalk";
+
+/**
+ * Loads the puppeteer-core revisions, aka recommended browser versions
+ */
+async function loadPuppeteerRevisions(): Promise<
+  { chrome: string } | undefined
+> {
+  try {
+    const revisions = await import(
+      "puppeteer-core/lib/cjs/puppeteer/revisions.js"
+    );
+    return revisions.PUPPETEER_REVISIONS;
+  } catch {
+    try {
+      const revisions = await import(
+        "puppeteer-core/lib/esm/puppeteer/revisions.js"
+      );
+      return revisions.PUPPETEER_REVISIONS;
+    } catch {
+      return undefined;
+    }
+  }
+}
+
+const INSTALL_TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
+const MAX_INSTALL_RETRIES = 3;
+
+/**
+ * Validates and sanitizes cache directory path
+ */
+function validateCacheDir(cacheDir: string | undefined): string | undefined {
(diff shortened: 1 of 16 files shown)


packages/common/src/browser-installer.ts
```diff
@@ -0,0 +1,187 @@
...
+ */
+export async function ensureBrowser(
+  browserType: Browser = Browser.CHROME,
+): Promise<string> {
+  // Try to get existing browser path first
+  try {
+    const browserPath = executablePath();
+    if (browserPath) {
+      // Verify the executable actually exists
+      const fs = await import("fs");
```
edoardopirovano, line 111: Any particular reason we can't just import this at the top with everything else?

packages/common/src/browser-installer.ts
```diff
@@ -0,0 +1,187 @@
...
+  } catch (error) {
+    if (
+      error instanceof Error &&
+      !error.message.includes("Could not find browser")
+    ) {
+      console.debug(chalk.gray(`Browser detection info: ${error.message}`));
+    }
+  }
+
+  // Detect platform
```
edoardopirovano, line 125: I don't like these pointless comments that AI agents leave.

packages/common/src/browser-installer.ts
```diff
@@ -0,0 +1,187 @@
...
+        " Download from https://www.google.com/chrome/",
+    );
+  }
+
+  const validatedCacheDir = validateCacheDir(process.env.PUPPETEER_CACHE_DIR);
+
+  // Try to use a Chrome version compatible with the replay bundles.
+  // We prioritize the version that the replay orchestrator expects to avoid
+  // version mismatch warnings during execution.
+  const buildId = "138.0.7204.0";
```
edoardopirovano, line 145: Hmm, I think we should download the default version that Puppeteer does so when we bump that Chrome gets bumped too. If we must hard code a build ID then it should be at the top of the file.

packages/replay-orchestrator-launcher/src/index.ts
```diff
@@ -58,42 +60,41 @@ export const executeScheduledTestRunChunk = async (
...
-  // the reference to this specific version of Chromium to the replay bundle, since
-  // the replay bundle may use a slightly different version of Puppeteer, which
-  // may have a slightly different default Chromium path/version.
-  return executablePath();
+const getChromiumExecutablePath = async (): Promise<string> => {
+  // Ensure browser is available using our shared browser installer.
+  // The browser installer will prioritize Chrome versions that are compatible with
+  // the replay bundles to minimize version mismatch warnings. If the preferred
+  // version is unavailable, it will fall back to puppeteer-core's compatible version.
+  return await ensureBrowser();
```
edoardopirovano, line 99: This function is identical to `ensureBrowser`, can we drop it and just use that?

edoardopirovano (approved):

packages/common/src/browser-installer.ts
narobertson42: imagine import slowdown pretty minimal, will remove

packages/common/src/browser-installer.ts
narobertson42: I tried with stable but ran into some issues, will investigate further and see if there's any other approach in the browsers api we can use

PR #948, opened 2025-09-24
ENG-1417: command to compute base if not available with preprocessing step

phreppo:


packages/cli/src/commands/run-all-tests-in-cloud/run-all-tests-in-cloud.command.ts
```diff
@@ -221,6 +237,69 @@ const handler: (options: Options) => Promise<void> = async ({
...
+      cloudReplayBaseTestRun = await getGitHubCloudReplayBaseTestRun({
+        client,
+        headCommitSha: commitSha,
+      });
+
+      testRun = cloudReplayBaseTestRun.baseTestRun;
+    }
+  } catch (error) {
+    if (isOutOfDateClientError(error)) {
+      throw new OutOfDateCLIError();
```
edoardopirovano, line 296: Why do we need to catch and handle this error in the helper function? I think the catch in the top-level handler for the command is sufficient and this extra try-catch isn't doing anything.

packages/cli/src/commands/prepare-for-meticulous-tests/prepare-for-meticulous-tests.command.ts
```diff
@@ -0,0 +1,103 @@
...
+  try {
+    const client = createClient({ apiToken: apiToken_ });
+
+    // Non-Github-hosted projects are currently not supported
+    const cloudReplayBaseTestRun = await getGitHubCloudReplayBaseTestRun({
+      client,
+      headCommitSha: headCommit,
+    });
+
+    if (cloudReplayBaseTestRun.baseTestRun === null) {
```
edoardopirovano, line 53: We also need to check that this is a PR run, otherwise we will start an infinite chain of runs going back through all their Git history and DDOS ourselves and their CI. You'll probably need to extend the backend endpoint to return whether the commit is a PR one or not.

packages/cli/package.json
```diff
@@ -44,6 +44,7 @@
     "chalk": "^4.1.2",
     "ci-info": "^4.0.0",
     "cli-progress": "^3.12.0",
+    "loglevel": "^1.9.2",
```
edoardopirovano, line 47: I think we should do 1.8.0 since that's what we used elsewhere. Also I guess you only imported this for a type declaration so it could be in `devDependencies` rather than `dependencies` which means it can be omitted from the final build. (But actually it doesn't matter because elsewhere we do properly depend on it so it will end up in the final build anyway.)

packages/cli/src/commands/run-all-tests-in-cloud/run-all-tests-in-cloud.command.ts
```diff
@@ -16,6 +21,9 @@ import {
   OutOfDateCLIError,
 } from "../../utils/out-of-date-client-error";
 
+const POLL_FOR_BASE_TEST_RUN_INTERVAL_MS = 10_000;
+const POLL_FOR_BASE_TEST_RUN_MAX_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes
```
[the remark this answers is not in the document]
phreppo: Solved in d116522d2acb8ed6adb46a993d1a7786cba2a04c

packages/cli/src/commands/run-all-tests-in-cloud/run-all-tests-in-cloud.command.ts
phreppo: Solved in d8d3bd712d30e53fdd43882fd7b013053e81c29d

packages/cli/package.json
phreppo: Solved in 0187e2ccd7fb8439dcc48785513304dfe199ea10

packages/cli/src/commands/prepare-for-meticulous-tests/prepare-for-meticulous-tests.command.ts
phreppo: Solved in fe3fcc9d4216e9f234cf40ce1f2060c366aa156a. Observe that this requires the updated backend API to be available first.

packages/cli/src/commands/run-all-tests-in-cloud/run-all-tests-in-cloud.command.ts
```diff
@@ -221,6 +260,61 @@ const handler: (options: Options) => Promise<void> = async ({
...
+  let lastTimeElapsed = 0;
+
+  while (!testRun) {
+    const timeElapsed = Date.now() - startTime;
+    if (timeElapsed > POLL_FOR_BASE_TEST_RUN_MAX_TIMEOUT_MS) {
+      logger.error(
+        `Timed out after ${POLL_FOR_BASE_TEST_RUN_MAX_TIMEOUT_MS / 1000} seconds waiting for base test run`,
+      );
+      // We cannot proceed without base: hard fail
+      process.exit(1);
```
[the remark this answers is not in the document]
phreppo: Solved in 1edd94d875e41118ccf2c3b9ea842e76ce611e35

